PR #36, opened 2016-10-31
Improve syntax highlighting and indentation behavior, plus postgresql pools

smartin015:


smartin015:
Pool was required here as we were previously failing to close opened postgres clients and causing our production instance to run out of resources (sockets or memory, unsure which).

PR #39, opened 2016-11-27
Add app preview support to editor

smartin015:
- Converted all styles from inline to SCSS
- Removed a bunch of now-defunct translation code (some still remains, will do a future cleanup sweep)
- "dirty" state is now "editor" state, to include the most recent XML render, the current cursor line, etc.
- Completely re-did the QDL->XML parsing code to follow a line-by-line approach. This gives us the ability to render errors and turn a text cursor position into a "selected block". See QDL.tsx's "renderXML" for usage. 
- Annotations are now shown in the Ace editor gutter when errors in QDL parsing occur.
- Nested the App redux state *inside* the existing redux state. Basically state.preview is viewed as "state" by the app.


app/components/Splash.tsx
```diff
@@ -50,23 +25,20 @@ const Splash = (props: any): JSX.Element => {
...
-            target="_blank"
-          >
+        <div className="learnMore">
+          <a href="https://github.com/Fabricate-IO/expedition-quest-ide/blob/master/docs/markdown_guide.md" target="_blank">
             learn more
           </a>
         </div>
-        <div style={{'background': '#141414'}}>
-          <img src="/assets/img/app-screenshot.png" style={styles.screenshot}/>
+        <div clasName="screenshot">
```
toddmedema: typo

toddmedema:
looks good to me, except for the one typo noted (should be className). After fixing, please double check that the splash screen works (ie in incognito window). After that, feel free to merge

app/components/Splash.tsx
toddmedema: Fixed!

toddmedema:
Usability note: would be much better if "play from the start" skipped the setup screen, annoying to have to click through that every time

toddmedema:
Also really thinking that (can totally be a separate push) some sort of "insert" shortcut menu for inserting choices, combat, etc would be really handy. When I'm in the writing mode, it's annoying to have the think about the "correct" way to input a fight, I just want to type / click "fight" and keep writing

toddmedema:
Bug: Looks like "unsaved changes" indicator is no longer hooked up to saving

PR #45, opened 2016-11-29
<Insert> tool in app bar

toddmedema:
Done so far:

- Styled UI dropdown that triggers a reducer, that inserts text at the provided line

Still to do:

- blocks and intelligent intendenation
  - problem: inserting something right after a choice should be indented, but the choice isn't indented
  - problem: inserting (really, any) node should leave / insert a blank line both below and above
  - problem: doesn't insert at the end of a block; better to add at the end of the block 
  - block.getResultAt
  - insert at the end of a block
  - find the first character of the last line of the block; include all of the spaces necessary to properly indent

- track indent position in textview?
  - same way we calculate state.line, re-calculate and store the magical indent number on each cursor change (should also catch moveCursorTo changes from insertions)

- test cases!
  - that interface w/ mdRealtime? Hmmmm....
  - needs to both fetch and set around mdRealtime
  - also, editor / UI changes?...
  - sounds like you need a modular logic-only function in the middle!

- move cursor to end of inserted text
  - how to get access to these function(s) from a reducer?
  `ace.moveCursorTo(Number row, Number column)` 
  `ace.navigateLineEnd()`

- Finally, fill in the actual text options
  - be as smart as possible about post-insertion text. For example, encounter should also include the first enemy list hyphen, choice should include the spaces to extra-indent the next line
    - might have to change the actual text generation to happen in the reducer, there's a lot of (state-dependent) logic here....


toddmedema:
Ditching this; too much has changed and there are more important / valuable things to work on

PR #49, opened 2016-12-03
Added op support for enemy lists

smartin015:
This allows for three formats of enemy within a combat node:

```
_combat_

- Skeleton Swordsman
- {{anEnemyExpression}}
- {{condition}} BadGuy
```

The first is a standard enemy, with no additional parsing.

The second evaluates the expression and uses the result value as the enemy.

The third conditionally shows the enemy based on the expression's evaluation result.


toddmedema:
Looks good! Only thought - do we want to keep support for enemies in the JSON list, even if it's not the publicly documented way? Seems like it'd be straightforward to support both (they're both lists at the end of the day)

smartin015:
It's already supported :) this doesn't remove JSON list assignment.

toddmedema:
Ah gotcha. Looks like the JSON test case was replaced by the list test case? (which is probs why I got confused) - might be worth keeping test cases for both

smartin015:
Now we have a separate test for setting enemies via JSON.

toddmedema:
Lgtm

PR #70, opened 2016-12-07
Add op debugging tools

smartin015:
https://github.com/Fabricate-IO/expedition-quest-creator/issues/51


toddmedema:
Part of adding ops: https://github.com/Fabricate-IO/expedition-quest-creator/issues/43

toddmedema:
Make sure to hit these points as part of the console: https://github.com/Fabricate-IO/expedition-quest-creator/issues/129 (or, at least as many as practical)

smartin015:
Finally ready for review - give 'er a pull and try it out!

app/components/App.tsx
```diff
@@ -26,12 +30,16 @@ const App = (props: AppProps): JSX.Element => {
...
-          onTouchTap={(event: any) => props.playFromStart(props.editor, props.quest)} />
+        <IconButton
+          onTouchTap={(event: any) => props.playFromCursor(props.scope, props.editor, props.quest)}
+          tooltip="Play from Cursor" >
+          <PlayIcon/>
+        </IconButton>
+        <IconButton
+          onTouchTap={(event: any) => props.playFromCursor({}, props.editor, props.quest)}
+          tooltip="Play from Cursor (blank context)" >
+          <ReplayIcon/>
```
toddmedema: For some reason, these two icons and their uses confuse me. Every time I look at them, I have to think "wait, does replay mean replay with my current state? Or does it not?"

Two possible solutions:

- switch back to text buttons, since it's a pretty complex difference that standard icon language doesn't really have a good verb for
- make the play button the only button, and have it play with a new state (since that's the most intuitive thing), but if you ctrl / cmd + click it, have it replay with current state (and mention that in the tooltip)

app/style.scss
```diff
@@ -131,3 +136,120 @@ html, body, #react-app {
     margin: auto;
   }
 }
+
```
toddmedema: Styling / UI feedback:

- should start hidden, with a "show context explorer" button floating on the bottom middle of the screen (regardless of screen height)
- when expanded, there should be a "hide context explorer" button sitting on top of it that hides it
- when shown, the context explorer should initially appear with a fixed height (for example, 100px) at the bottom of the screen, regardless of viewport height
- the horizontal title bar separating the context explorer from the rest of the editor should be more visually prominent (right now, you have to squint to see it). Making it more prominent would also eliminate the need for the ` - ` emphasizers on the section headers `initial context` and `context history`
- speaking of section titles, would make it more consistent with the UI to use the same font as the rest of the UI buttons

smartin015: These should all be done.

app/components/App.tsx
smartin015: Done (text only)

smartin015:
Additionally:

* Fixed webpack to re-allow hot loading
* Added some test stubs
* Moved required new components into the `./base` folder

toddmedema:
@smartin015 Feedback:
- when the context explorer is hidden, the whole bottom bar should be clickable to show it
- on the font front... love the title in Cinzel, but seeing the other buttons in a serif font just looks wrong, we should switch the buttons (and all text in the context explorer window) back to their previous font
- initial context: {} <- should be wrapped in double brackets {{}}
- padding: most of the elements need more padding - aka anything that's touching the edge of its box ("Context Explorer", "+/-", "Initial Context: {{ / }}", "Empty Scope History"
  - a handy way to determine how much padding you should give them: line them up with other elements. For example, the +/- should be vertically aligned with the login icon above it
- buttons:
  - I feel like playing with a blank context should be the default. Although it might be easier to preserve context by default for advanced writers, this could lead to a lot of confusion and unexpected behavior for new writers (IMHO less confusion > slightly faster, you can always type a bit more if needed, but if you don't understand what's going on, you're more likely to get frustrated and give up entirely)
  - the other button (now "Play from Cursor (preserve context)" or similar) should only appear if the context explorer is open. Keeps it simpler for the average writer, plus, preserving context is a dangerous idea w/o the explorer open (ie you don't know what you're preserving)

smartin015:
* Done
* Done
* Done
* Done
* Done and done.

Done!

toddmedema:
Nice! More feedback (sorry):

- spacing
  - seems to be too much space below the bottom / closing double brackets (also too much space above them / between them and the textarea - should match the space between "initial context" and the text area)
  - can we line up "context explorer", "initial context", "}}" and the text area to match the same left padding as the "new" button on the app bar (and update the right spacing on the right side of the text area to match) - still feels a bit haphazard / too many different positions
  - the spacing on the "<" to set initial state from the current state seems kinda arbitrary. Can we either vertically center it to each list item (which might not be a great idea with long contexts) or align it to the bottom of the element? (which'd work better given that the most recent one is on the bottom)
  - the "<" also feels like there's uneven spacing on the left and right, could serve to shrink up the spacing on the left. Note that the hovertext seems to get hidden behind the other half of the explorer, we might need to boost its z-index
- alright, clicking the whole bar is somewhat weird and annoying / easy to accidentally click. Also, there's currently no indication that it's clickable (ie no hover effect and no cursor change). Can we instead make just the "context explorer" and "+" text clickable, but make sure they both have hover effects (cursor change and background color change, like the other UI buttons)
  - clicking on "Context explorer" should also close it when open
- it's weird that the "Context explorer" title is actually less prominent that the "initial context" right below it - seems like the wrong hierarchy. Can we:
  - switch both to use the sans-serif font that's used everywhere else (think they + the closing "}}" are the only things left using that font, then we can un-import it)
  - make "context explorer" full white, and make "initial context {{" and "}}" a the faded grey that "context explorer" is currently
- text area: can you make the help text "# Write mathjs here..." hint text? http://www.material-ui.com/#/components/text-field

app/components/Main.tsx
```diff
@@ -6,8 +6,7 @@ import SplashContainer from './SplashContainer';
 import QuestAppBarContainer from './QuestAppBarContainer';
 import QuestIDEContainer from './QuestIDEContainer';
 import ContextEditorContainer from './ContextEditorContainer';
-import RemoveIcon from 'material-ui/svg-icons/content/remove'
-import AddIcon from 'material-ui/svg-icons/content/add'
+import FlatButton from 'material-ui/FlatButton';
```
toddmedema: nit: no semicolons on import

PR #83, opened 2016-12-09
doc revamp

toddmedema:
- update the example Mistress Malaise and Oust Albanus quests
- updated our response time estimates on 404 page
- total revamp of `qdl_guide` - uses new syntaxes, and streamlines to be more of a quick reference, since the starter quest now serves as the more story-driven / thematic introduction to QDL
- updated the index page to provide more context

NOTE: once this all looks good, I'll go back and add a table of contents of links to the top of the new qdl guide


docs/quest_spec.md
```diff
@@ -1,13 +1,13 @@
 Quest Specification
 ===================
 
-Quest files are used to build custom roleplaying scenarios with the Expedition App in a custom _Question Definition Language_.
```
smartin015: Looks like we're making the Expedition Question Creator....

PR #84, opened 2016-12-09
Improve QDL error display and documentation

smartin015:
Added error codes for all errors in the editor, and added error doc templates. Errors are now displayed more compactly in the IDE.

This also has a small functional change to treat triggers as always single-line entities, i.e. a headerless roleplay card after a trigger is valid.


app/parsing/QDLParser.tsx
```diff
@@ -264,7 +268,7 @@ export class QDLParser {
       if (blocks.length !== 1) {
         log.err(
           'quest block group cannot contain multiple blocks',
```
toddmedema: why are internal logs above on a single line, but this one on multiple lines?

also, thought: putting the error code before the message would make these more readable

app/parsing/render/BlockRenderer.tsx
```diff
@@ -239,16 +233,16 @@ export class BlockRenderer {
     // Parse meta using the block itself.
     // Metadata format is standard across all renderers.
     if (!block) {
-      if (log) log.err('Missing quest root block', '404');
+      if (log) log.err('missing quest root block', '419');
```
toddmedema: thought: rather than checking for if (log) on every log (seems bug-prone), could set a default at the top of the function, something like log = log || { err: console.log }

toddmedema:
Note: we should still plan on finishing up the recommended code changes here, but merging into master and deploying for the sake of having the user-facing side live on launch

PR #85, opened 2016-12-09
Endparse

toddmedema:
Fixes the 503 error always happening when there's an `**end**` element at the end of a quest (because it was inserting a blank card after all triggers).

Updated test case.

Rolling out to dev (and prod) if everything looks good, but still worth reviewing


smartin015:
LGTM

PR #87, opened 2016-12-09
cleanup

toddmedema:
- `/quests/model` -> `/model/quests`
- `updated minimum node requirement to v6, which is a long term supported version, but also several years newer :)
- updated code to use `const` instead of `var` when possible (following modern JS recommendation)
- removed unused gulp file
- removed some unused dependencies from package.json

Tested locally w/ a full node_modules wipe and install, and tested on dev - looks good. Merging to master, but worth reviewing eventually.


models/user.js
```diff
@@ -0,0 +1,8 @@
+/*
```
smartin015: What is this used for?

toddmedema: The user model? Keeps track of the last time the user logged in. Could expand it into other things, such as keeping track of quests for the file browser

smartin015: Gotcha, makes more sense after reading the other PRs.

PR #90, opened 2016-12-10
Model cleanup

toddmedema:
Update quests model to use the wombo-combo of Squel (for query building) + Joi (for validation) + some handy utils (for abstraction).

Note that you'll need to run `npm install`

Merging for progress, but worth reviewing


models/quest.js
```diff
@@ -18,170 +18,219 @@ CREATE TABLE quests (
...
+});
 
-function search(userId, params, cb) {
-  if (!params) {
-    cb(new Error("Search params not given"));
-  }
 
-  var filter_query = 'SELECT * FROM quests WHERE tombstone IS NULL';
-  var filter_params = {};
+exports.getById = function(id, cb) {
```
smartin015: It looks like this can get a quest regardless of its tombstone or publish state... I wonder if we should add some checks here so everyone can't read everything.

models/utils.js
```diff
@@ -0,0 +1,139 @@
+const Moment = require('moment');
```
smartin015: "utils" as a filename is the bane of readable code (it's essentially "some things that do stuff", which is literally what all code is).

I'd suggest splitting this into two separate files, maybe something like "generate_sql" and "js_pg_translation".

models/utils.js
```diff
@@ -0,0 +1,139 @@
...
+  dateKeys: ['created', 'published', 'tombstone', 'updated'], // columns that're always dates
+  regexJsToPg: /([A-Z])/g,
+  regexPgToJs: /_([a-z])/g,
+  replacerJsToPg: (match, p1) => '_' + p1.toLowerCase(),
+  replacerPgToJs: (match, p1) => p1.toUpperCase(),
+};
+
+
+// returns SQL string to delete from table based on filters provided
+exports.generateDeleteSquel = function (table, filters) {
```
smartin015: Technically this is generating SQL, not Squel.

models/utils.js
toddmedema: Given your suggestion [here](https://github.com/Fabricate-IO/expedition-quest-creator/pull/95#pullrequestreview-14419003) to privatize the js_pg_translation code, feel like it might make sense to turn utils into `runQuery`, which generates and runs queries and uses the translation code privately - working on a PR now

models/quest.js
toddmedema: Hmm. There is a tombstone check on search, but not on getById. So, people shouldn't be able to discover unpublished quests, which seems like a good starting point at least

models/utils.js
toddmedema: Good catch

PR #92, opened 2016-12-10
create user model

toddmedema:
Create user model with id, email, created, tombstone. Upsert user on login / auth.

Going to do some final testing on dev, then roll out on prod, but still worth post-reviewing


smartin015:
LGTM

PR #95, opened 2016-12-13
DB fixes

toddmedema:
- correctly set user.lastLogin to the time of last login, not when the server started
- streamline querying code


toddmedema:
Merging for the sake of progress; worth reviewing in the future (esp the weirdness w/ date validation defaults)

models/quest.js
```diff
@@ -99,11 +102,11 @@ exports.getById = function(id, cb) {
       if (results.length !== 1) {
         return cb({
           code: 404,
-          message: 'Not found'
+          message: 'Not found',
         });
       }
 
-      cb(null, Utils.objectPgToJs(results[0]));
```
smartin015: Looks like all the PgToJs/JsToPg are no longer used in quest.js or user.js. We should unexport them to keep the interface clean.

models/user.js
```diff
@@ -30,7 +29,6 @@ const poolConfig = {
   ssl: true,
 };
 const pool = new Pg.Pool(poolConfig);
```
smartin015: It's kind of weird that we have two query pools. Since all we're doing is passing them into utils.js (which I still think should be renamed, lol) we might as well remove them and add a single pool in utils.js. Means one less param to pass!

models/utils.js
```diff
@@ -11,12 +11,37 @@ const internals = {
...
+exports.queryOne = function (pool, query, callback) {
+  exports.query(pool, query, (err, result) => {
+    return callback(err, result[0]);
+  });
+};
+
+
 // returns SQL string to delete from table based on filters provided
+// Note: Postgres child objects with ON DELETE trigger references
+  // will also be deleted when parent is deleted
```
smartin015: Nit: extra indent

toddmedema: @smartin015 so, I'd been using an extra indent in comments to indicate a continuation of the same thought, vs a new line. IE:

```
// This thought is really 
  // long and here's more detail
// Here's another thought entirely
```

That being said, that's just a stylistic quirk of mine, so open to better ways to handle that :)

models/quest.js
toddmedema: Good point

models/utils.js
smartin015: I usually use newlines for that, e.g.

```
// This thought is really 
// long and here's more detail
//
// Here's another thought entirely
```

But really, we should just go with whatever comment style is more common so other folks can read our code more easily. Maybe this could be helped with #106.

models/user.js
toddmedema: Good idea!

PR #96, opened 2016-12-14
Comments

toddmedema:
Complete fix for comments, allowing them to be placed anywhere without issue - by removing them from the MD entirely before parsing begins! plus test case.

Rolling out for progress, but still worth reviewing


smartin015:
LGTM

PR #97, opened 2016-12-14
Continuous deployment

toddmedema:
Continuously deploy w/ Travis. The code now gets built on Heroku, so we're able to remove the build files from the repo, and no longer have to remember to separately build before pushing


smartin015:
Woah, spiffy. And a major cleanup too! :1st_place_medal:

PR #98, opened 2016-12-14
throw error(s) when attempting to publish default quest metadata

toddmedema:


toddmedema:
@smartin015 That'd be good. Seems like all of the action tests are just stubs right now. Any experience with setting up tests for actions?

smartin015:
Looks like mock-redux-store (and possibly nock if http requests are used) is recommended for unit testing async redux actions.

https://github.com/reactjs/redux/blob/master/docs/recipes/WritingTests.md#async-action-creators

PR #99, opened 2016-12-14
fix silent bug with untitled roleplaying after combat

toddmedema:
closes https://github.com/Fabricate-IO/expedition-quest-creator/issues/75


app/parsing/render/BlockRenderer.tsx
```diff
@@ -183,9 +182,9 @@ export class BlockRenderer {
         // We should only ever see event blocks within the combat block.
         // These blocks are only single lines.
         var extractedEvent = Object.assign({}, this.extractBulleted(line), {event: []});
-        if (!extractedEvent.text) {
+        if (extractedEvent == null || !extractedEvent.text) {
           log.err(
-            "lines within combat block must be event bullets or enemies; instead found \""+line+"\"",
+            "lines within combat block must be events or enemies;\ninstead found \""+line+"\"",
```
smartin015: Need to update the markdown doc with the new error text

toddmedema: @smartin015 that should be taken care of in the new PR that auto-generates all of the errors / docs :)

smartin015: Nice :D

PR #100, opened 2016-12-15
add support for inline icons

toddmedema:
See https://github.com/Fabricate-IO/expedition-app/pull/156


docs/icons.md
```diff
@@ -0,0 +1,23 @@
+# Expedition Quest Creator Icons
+
+You can insert inline icons into your quest using `[icon_name]`. Inline icons are currently supported in roleplaying cards, instruction elements and in the titles of choices
+
+At this time, you can use the following Expedition icons:
+
+* d20
```
smartin015: It'd be sweet if these also displayed the icons so people could see them right away.

toddmedema: It would be sweet.... ;) https://github.com/Fabricate-IO/expedition-quest-creator/blob/master/docs/icons.md

PR #103, opened 2016-12-19
Programmatic errors

### LINKED ISSUE #16: Improved + additional unit tests
Besides generally adding more tests, we can also get fancy and combine the error documentation + error test cases. Right now, we have to duplicate "good / doesn't throw error X" and "bad / throws error X" examples across documentation and tests, but would save significant maintenance effort (+ result in better tests + documentation) to have them specified in the same file + auto-generated

toddmedema:
TODO: 
- review breaking tests:
  - 412 - can't reproduce, triggers 414 and 417 if just _combat_, but not 412
  - 413 - can't reproduce, if it's `_combat_`, it's unbroken, anything else is just a roleplay title
  - 415 - can't reproduce, seems redundant with 416
  - 419 - can't reproduce, seems to only be possible w/ purposefully broken XML
  - 425 - can't reproduce, every data type works as a string

DONE:
- fix broken table of contents error links
- combine error documentation with test cases to eliminate redundancy and improve maintainability and quality of documentation.
  - create error definition files with examples of broken and fixed cases
  - auto-generate test cases against all broken and fixed test cases
    - removed simple error test cases from other tests (ie if a test case is checking for a specific, single error, makes more sense to live in the error definition file)
  - generate error doc markdown files based on error definitions; automatically generated on each test run (quite fast) so the docs are always up to date

closes #16


toddmedema:
@smartin015 could use your review on the breaking tests / how we can build test cases to confirm them (or, if it's impossible to reproduce, we can clean up the code!)

smartin015:
412: pretty sure this one can be removed.

413: This should fire on a bad JSON annotation, e.g:
`_combat_ {invalid_json}`

415: Try having an indented block inside combat:
```
_combat_

* Enemy1

  This block is indented for no reason
```

419: Safely guarded against with other errors, but we still need the return case when calling it outside of the parser (e.g. to show the current quest title in the top header bar). Suggest just removing the logging bit and the error code from docs.

425: Derp. Yep. Although we might later want to enforce some form of non-empty validation, we can clear this error for now.

lib/oauth2.js
```diff
@@ -75,6 +77,7 @@ router.post('/auth/google', // LOGIN
   },
   Passport.authenticate('google-id-token'),
   (req, res) => {
+console.log(req.get('origin'));
```
smartin015: Looks like ya got some console logs there

toddmedema: Derp derp

toddmedema:
@smartin015 alright, this should be caught up with all of the error-related feedback from the PR's - LMK if you see anything else, or if it looks good to go!

app/parsing/QDLParser.test.tsx
```diff
@@ -58,14 +58,6 @@ describe('QDLParser', () => {
     expect(prettifyMsgs(qdl.getFinalizedLogs()['error'])).toEqual(TestData.noHeaderError);
   });
 
-  it('errors if combat contains an invalid block', () => {
```
smartin015: Why was this removed? It looks still valid (error 416)

app/parsing/render/BlockRenderer.test.tsx
```diff
@@ -377,35 +377,6 @@ URL: 428`);
       expect(prettifyHTML(block.render + '')).toEqual('<quest title="Quest Title" author="Test" maxplayers="2"></quest>');
       expect(prettifyMsgs(log.finalize())).toEqual(TestData.badParseQuestAttrError);
     });
-
-
-    it('errors if unknown quest attribute', () => {
```
smartin015: Same question for these tests (Errors 426 and 427)

app/parsing/validation/Normalize.tsx
```diff
@@ -33,7 +33,7 @@ export class AttributeNormalizer {
     }
 
     if (this.log) {
-      this.log.err(k + " should be a string, but is "+typeof(v), '425');
+      this.log.err(k + " should be a string, but is " + typeof(v), '425');
```
smartin015: The markdown docs for err 425 were removed, but the error still exists...

docs/index.md
```diff
@@ -42,6 +42,8 @@ Quest errors:
 * [425 <key> should be a string, but is <type>](errors/425.md)
```
smartin015: Should verify these are all correct (412, 419, and 425 aren't necessary anymore)

errors/errors.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {prettifyMsgs} from '../app/parsing/Logger'
+import {BlockList} from '../app/parsing/block/BlockList'
+import {XMLRenderer} from '../app/parsing/render/XMLRenderer'
+
+const Errors: any = require('./errors');
+const expect: any = require('expect');
+const DocGenerator: any = require('./docGenerator');
+
+// Re-generate the docs whenever you run the tests
+DocGenerator.generate();
```
smartin015: I like the idea of auto-generating the docs, but this seems way outside the scope of this file. Can you move the doc generation stuff to a separate file (preferably without a .test.tsx extension)?

app/parsing/QDLParser.test.tsx
toddmedema: @smartin015 to clean up the code, I removed duplicate tests - since the error definitions auto-test that specific error, I removed tests that existed elsewhere that tested against a singular error code

app/parsing/validation/Normalize.tsx
toddmedema: Good catch

docs/index.md
toddmedema: Another good catch :) #codingWhileRelativesAreRunningAround

errors/errors.test.tsx
toddmedema: Hmm.

So the original idea was that rebuilding the docs when tests are run would keep the docs always up to date (rather than having to remember to run a separate build step)

Does that premise seem reasonable? (aka clarifying: is this query just about putting it in a separate file that also gets executed on `npm test`? Or is it outside the scope of running tests?)

app/parsing/QDLParser.test.tsx
smartin015: Ah, carry on then :) Just wanted to make sure we weren't losing coverage.

errors/errors.test.tsx
smartin015: I agree with regen during development. I'm alright with it running as part of tests, although it does stretch the premise of "tests" pretty hard. It'd be even better to make it part of the main ide/webpack dev script, if possible. 

But regardless of if that's possible, it should definitely be its own file so we're not giving side effects to our unit tests.

toddmedema: Fair :) Updated!

(It'd also make sense to make it part of the webpack script.... except that I offloaded that to happen on the server / after deploy b/c I kept forgetting to run it locally lol)

smartin015:
Looks good - SHIP SHIP SHIP

PR #104, opened 2016-12-20
"publish button" changes to "validation errors" when errors present

toddmedema:
rolling out for progress, still worth reviewing


app/components/QuestAppBar.tsx
```diff
@@ -48,10 +49,12 @@ const QuestAppBar = (props: QuestAppBarProps): JSX.Element => {
...
-  const publishButton = (props.quest.valid === false) ? // default to showing the is valid button
+  const errors = props.annotations.filter((annotation) => { return annotation.type === 'error' });
+  const errorLabel = (errors.length > 1) ? 'Validation Errors' : 'Validation Error';
+  const publishButton = (errors.length > 0) ?
     <FlatButton
       style={styles.button}
-      label="Validation Error(s)"
+      label={errorLabel}
       icon={<AlertError />}
       onTouchTap={(event: any) => props.onMenuSelect('PUBLISH_QUEST', props.quest)} />
```
smartin015: In hindsight, this feels kind of weird. Why should clicking a button labeled "validation errors" publish your quest? Maybe we should make this disabled until the user fixes the errors, or instead indicate the validation errors with a model that asks if the user *really* wants to submit with them.

toddmedema: Hmm. I was thinking it might be most useful to have clicking it scroll the editor to the first error line? (maybe flash that line too? in case that line is already in view)

smartin015: That makes sense... sounds like some decent work though. Maybe we can make it disabled for now, and file a FR?

toddmedema: Added FR: https://github.com/Fabricate-IO/expedition-quest-creator/issues/107
Will add a PR to disable soon (likely tonight)

toddmedema: @smartin015 PR to disable button: https://github.com/Fabricate-IO/expedition-quest-creator/pull/110

PR #108, opened 2016-12-27
Utils.js -> Query.js

toddmedema:
Because a code file named utils is nonsense, and a bunch of other good-idea cleanups around that code courtesy of @smartin015


models/query.js
```diff
@@ -0,0 +1,207 @@
...
+};
+
+
+/*
+ * deletes all rows in table that match filters
+ *
+ * Note: Postgres child objects with ON DELETE trigger references
+ * will also be deleted when parent is deleted
+ */
+exports.delete = function (table, filters, callback) {
```
smartin015: This is risky... is it used anywhere? It's far safer to tombstone everything and garbage collect once we start running into space issues.

toddmedema: Good call, updated (turned out .delete wasn't used anywhere, so changed it to .deleteId, and replaced the quest model's hand-coded delete function with that)

models/query.js
```diff
@@ -66,25 +66,16 @@ exports.getId = function (table, id, callback) {
 
 
 /*
- * deletes all rows in table that match filters
+ * deletes row with matching id from table. Returns the id deleted.
  *
  * Note: Postgres child objects with ON DELETE trigger references
  * will also be deleted when parent is deleted
```
smartin015: Nit: Get rid of postgres note, mention we're actually tombstoning and not deleting.

PR #112, opened 2016-12-29
More icons!

toddmedema:


docs/icons.md
```diff
@@ -4,36 +4,74 @@ You can insert inline icons into your quest using `[icon_name]`. Inline icons ar
 
 We recommend only using icons as unique symbols, not as a replacement for words. (i.e. **DON'T:** use the dice icon instead of saying "roll a dice". **DO:** use the bandit icon to show what the thief's tatoo looks like)
 
-At this time, you can use the following Expedition icons:
+At this time, you can use the following Expedition-themed icons:
 
-* d20 / roll <img src="https://cdn.rawgit.com/Fabricate-IO/expedition-app/master/app/images/d20_small.svg" width="48" height="48"/>
+<img src="https://cdn.rawgit.com/Fabricate-IO/expedition-app/master/app/images/d20_small.svg" width="48" height="48"/> d20 / roll
```
smartin015: I spy another potentially auto-generatable page... :)

toddmedema: Was thinking about that :-) the challenge / annoying thing is that icons are spread across two repos, but I suppose for auto-generating here you could just reach down into the Expedition-app node module. 

Challenge: wonder if there's a better file / naming structure for the icons since now most of them are only going to have "small" versions. Think the large versions would only be used for background images? But, for icons that we aren't custom making, it'd be weird and annoying to have to make two, three or even four copies of the same file (large and small, black and white) 

Also challenging / brainstorming: if this is auto generated, what's the best way to store the icons credit link?

smartin015: Credit link sounds like we'd need some kind of metadata JSON or text file, unless we want to annotate the SVGs themselves with metadata (https://www.w3.org/TR/SVG/metadata.html)

As discussed, naming is hard and has tradeoffs.

Should also limit the number of icons we're publishing, to prevent icon bloat.

PR #118, opened 2017-01-07
Bugfix: allow single quotes in metadata / SQL

toddmedema:


models/query.js
```diff
@@ -90,7 +96,9 @@ exports.create = function (table, object, callback) {
 // inserts object into table, updating it if ID already exists
 exports.upsert = function (table, object, callback) {
 
-  object = objectJsToPg(object);
+  object = internals.objectJsToPg(object);
+
+console.log(object);
```
smartin015: This looks like leftover logging (also see "SQL command" console log)

toddmedema: Done. Re: SQL command console log, thought it might be worth keeping in for much faster debugging when things do go wrong, since the first thing I always end up looking at w/ publishing issues is "what's the raw SQL being run?"

Having recently executed commands can also be really helpful for re-playing issues. Might even be worth upgrading the SQL command logging to log to a file?

smartin015:
From offline conversation: keeping SQL makes sense to me. Console's probably the best place for it.

PR #134, opened 2017-01-08
record UI interactions and errors as analytics events

toddmedema:


smartin015:
LGTM. And surprisingly simple.

PR #141, opened 2017-01-12
fix multi auth bug

toddmedema:
- use a fork of the auth lib that allows multiple keys
- requires that the iOS key be added (comma separated to config.json client id)


lib/oauth2.js
```diff
@@ -13,7 +13,7 @@ const User = require('../models/user');
 // object, which will be set at `req.user` in route handlers after
 // authentication.
 Passport.use(new GoogleTokenStrategy({
-    clientID: Config.get('OAUTH2_CLIENT_ID'),
+    clientID: Config.get('OAUTH2_CLIENT_ID').split(','), // allow for multiple keys
```
smartin015: I feel like we could've just had two instances of Passport, instead of forking the repo...

toddmedema: Thought about that, but then I saw the line `Passport.authenticate('google-id-token'),` and realized that the plugin name's defined in the plugin's code, so there didn't seem to be a straightforward way to make two unique instances of it? Might also just submit the multi-id change to the main repo as a PR, so that we can stop using a fork

toddmedema: https://github.com/jmreyes/passport-google-id-token/pull/13

PR #152, opened 2017-01-18
add node mail feedback endpoint

toddmedema:
Pairs with https://github.com/Fabricate-IO/expedition-app/pull/184


routes.js
```diff
@@ -1,36 +1,68 @@
...
     state: JSON.stringify(res.locals),
   });
 });
 
 
+// TODO validate with hapi: require title, author_name, author_email, feedback, players, difficulty
+router.post('/feedback', (req, res) => {
+  res.header('Access-Control-Allow-Origin', req.get('origin'));
+  const params = JSON.parse(req.body);
+  const message = `Some adventurers have sent you feedback on your quest, hurray!
```
smartin015: hurray => hooray

routes.js
```diff
@@ -1,36 +1,68 @@
...
+  \r\n\r\n
+  They played with ${params.players} adventurers on ${params.difficulty} difficulty.
+  \r\n\r\n
+  Their feedback:
+  \r\n\r\n
+  "${params.feedback}"
+  \r\n\r\n
+  If you have any questions about the quest creator, or run into any bugs, you can email Expedition@Fabricate.io
```
smartin015: If you're sending as HTML... this could be linkified.

routes.js
smartin015: It looks like most of this text shows up as a quotebox in my gmail...

routes.js
```diff
@@ -40,25 +72,20 @@ router.post('/quests', (req, res) => {
...
-      if (err) {
-        res.header('Access-Control-Allow-Origin', req.get('origin'));
-        res.header('Access-Control-Allow-Credentials', 'true');
-        console.log(err);
-        return res.status(500).end("Search Error");
-      }
-      result = {error: err, quests: quests, nextToken: nextToken};
-      console.log("Found " + quests.length + " quests for user " + res.locals.id);
+  const params = req.body;
+  quests.search(res.locals.id, params, (err, quests, nextToken) => {
```
smartin015: Removing the try/catch could cause server task failure (due to internal exception), which would de-auth anyone using the quest IDE. Why was it removed?

routes.js
toddmedema: good idea

routes.js
toddmedema: fair. Got overzealous in code cleanup

routes.js
toddmedema: Just tried on my end, and the first email didn't have the collapsed sections. Must be b/c you had a couple

routes.js
```diff
@@ -1,47 +1,78 @@
...
+  <p>They played with ${params.players} adventurers on ${params.difficulty} difficulty.</p>
+  <p>Their feedback:</p>
+  <p>${htmlFeedback}</p>
+  <p>If you have any questions about the quest creator, or run into any bugs, you can email <a href="mailto:Expedition@Fabricate.io">Expedition@Fabricate.io</a></p>
+  <p>Happy Adventuring!</p>
+  <p>Todd & Scott</p>`;
+  // turn end of paragraphs into double newlines
+  const textMessage = htmlMessage.replace(/<\/p>/g, '\r\n\r\n').replace(HTML_REGEX, '');
+
+  Mail.send(params.authorEmail, 'Feedback on your Expedition quest: ' + params.title, textMessage, htmlMessage, (err, result) => {
```
smartin015: Could the title contain HTML renderable by email clients? I'm not sure, but probably better to filter it anyways.

PR #159, opened 2017-01-28
code folding!

toddmedema:
Closes https://github.com/Fabricate-IO/expedition-quest-creator/issues/113


app/components/base/QDLMode.tsx
```diff
@@ -25,40 +27,111 @@ var QDLHighlightRules: any = function() {
...
-export var QDLMode: any = function() {
-    // set everything up
-    this.HighlightRules = QDLHighlightRules;
-    this.$outdent = new MatchingBraceOutdent();
-    // TODO: Code folding
-    //this.foldingRules = new MyNewFoldMode();
+var QDLFoldMode: any = function() {};
+oop.inherits(QDLFoldMode, BaseFoldMode);
+
+(function() {
```
smartin015: This seems kind of roundabout/oldstyle... could this be an ES6 class instead?

app/components/base/TextView.tsx
```diff
@@ -84,8 +85,11 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
       ref.editor.$blockScrolling = Infinity;
 
-      // Set our custom mode.
-      ref.editor.getSession().$mode = mode;
+      // Set our custom mode. DUCT TAPE!!!
+      session.$mode = mode;
+      session.$foldMode = mode.foldingRules;
+      session.getFoldWidget = function(row: number) { return mode.foldingRules.getFoldWidget(session, '', row); };
+      session.getFoldWidgetRange = function(row: number) { return mode.foldingRules.getFoldWidgetRange(session, '', row); };
       ref.editor.getSession().bgTokenizer.setTokenizer(mode.getTokenizer());
```
smartin015: Might as well use the session var here, too

package.json
```diff
@@ -37,7 +37,7 @@
     "brace": "^0.8.0",
     "cheerio": "^0.20.0",
     "copy-webpack-plugin": "^4.0.0",
-    "expedition-app": "https://github.com/Fabricate-IO/expedition-app.git#375be171fb42533f645105c910e141670aa9327e",
+    "expedition-app": "https://github.com/Fabricate-IO/expedition-app.git#732d3b339afbbe1f6375187a76b04e7e5ba0fefb",
```
smartin015: Is this required for code folding, or just a periodic update?

app/components/base/QDLMode.tsx
```diff
@@ -25,40 +27,111 @@ var QDLHighlightRules: any = function() {
 };
 oop.inherits(QDLHighlightRules, MarkdownHighlightRules);
 
-export var QDLMode: any = function() {
-    // set everything up
-    this.HighlightRules = QDLHighlightRules;
-    this.$outdent = new MatchingBraceOutdent();
-    // TODO: Code folding
-    //this.foldingRules = new MyNewFoldMode();
+var QDLFoldMode: any = function() {};
```
smartin015: I could see this benefitting from some unit tests...

package.json
toddmedema: periodic update

app/components/base/TextView.tsx
toddmedema: Yep

app/components/base/QDLMode.tsx
toddmedema: Added some stubs; looks like Brace / Ace doesn't work outside of the browser, so we'd have to stub several parts of it (at some point, IMHO, the fragility of stubbing an entirely library reduces the value of tests; we can discuss if that's the case here)

app/components/base/QDLMode.tsx
toddmedema: Just tried, but hit a wall with a weird TypeScript error around `Type 'any' is not a constructor function type.`

Unless there's a big advantage to switching syntaxes, inclined to leave it (only seemed to save 1-2 LOC when I tried switching)

app/components/base/QDLMode.tsx
```diff
@@ -25,40 +27,111 @@ var QDLHighlightRules: any = function() {
...
+    };
+
+    this.getFoldWidgetRange = function(session: any, foldStyle: any, row: number): any {
+
+        var line = session.getLine(row);
+
+        if (!line.match(this.foldingStartMarker))
+          return;
+
+        function getIndent(line: string): number {
```
smartin015: This and getImportance could be private methods in ES6-land

app/components/base/QDLMode.tsx
```diff
@@ -25,40 +27,111 @@ var QDLHighlightRules: any = function() {
...
+          return -1;
+        }
+
+        const startRow = row;
+        const startColumn = line.length;
+        const maxRow = session.getLength();
+        const startIndent = getIndent(line);
+        const startImportance = getImportance(line);
+        var endRow = maxRow;
+        var endColumn = 9999;
```
smartin015: Is this a normal practice? There's no "unbounded end column"?

app/components/base/QDLMode.tsx
smartin015: Consistency :) want to try pair-debugging it?

app/components/base/QDLMode.tsx
smartin015: I was thinking more that we test QDLFoldMode specifically (without the rest of Ace). To test the bulk of it (i.e. `getFoldWidgetRange`) we'd just need to make a mock `session` with `getLine()` and `getLength()`.

smartin015: Also... I don't think the stubs were added.

app/components/base/QDLMode.tsx
toddmedema: Updated to use the length of the endRow

app/components/base/QDLMode.tsx
toddmedema: Updated

app/components/base/QDLMode.tsx
toddmedema: Woops, added stubs.

Right, the other problem is that just importing it breaks ace, since window doesn't exist. Not sure how to fix that one, since importing anything that uses brace then tries to import window. We could split out the regexes, but the rest of the file is heavily dependent on brace (even `getFoldWidgetRange`, which uses `brace.Range`)

app/components/base/QDLMode.tsx
```diff
@@ -25,40 +28,105 @@ var QDLHighlightRules: any = function() {
...
-export var QDLMode: any = function() {
-    // set everything up
-    this.HighlightRules = QDLHighlightRules;
-    this.$outdent = new MatchingBraceOutdent();
-    // TODO: Code folding
-    //this.foldingRules = new MyNewFoldMode();
 
+class QDLFoldMode {
+
+  static foldingStartMarker = /(^\s*)((# .*)|(\* .*)|(_.*_))/; // all possible markers, combination of foldingStartMarkers
```
smartin015: There might be a better way to compose foldingStartMarkers to prevent drift: http://stackoverflow.com/a/41870726

app/components/base/QDLMode.tsx
```diff
@@ -25,40 +28,105 @@ var QDLHighlightRules: any = function() {
...
+  static foldingStartMarkers = [ // most to least important
+    /(^\s*)(\* .*)/, // * choices
+    /(^\s*)(_.*_)/, // _cards_
+    /(^\s*)(# .*)/ // # titles
+  ];
+
+  static _getIndent(line: string): number {
```
smartin015: Ditch the underscores, and use `private` instead

app/components/base/QDLMode.tsx
```diff
@@ -25,40 +28,105 @@ var QDLHighlightRules: any = function() {
...
+    let indent = 0;
+    while (line[indent] === ' ') {
+      indent++;
+    }
+    return indent;
+  }
+
+  static _getImportance(line: string): number {
+    // check against most important first
```
smartin015: What does important mean? Does it fold around the other folds?

app/components/base/QDLMode.tsx
```diff
@@ -25,40 +28,105 @@ var QDLHighlightRules: any = function() {
...
+    }
+    return -1;
+  }
+
+  getFoldWidget(session: any, foldStyle: any, row: number) : string {
+      var line = session.getLine(row);
+      if (!QDLFoldMode.foldingStartMarker.test(line))
+          return "";
+
+      if (line[0] == "`") {
```
smartin015: What does this mean? Is this a manual fold?

app/components/base/QDLMode.tsx
toddmedema: Good idea, done

app/components/base/QDLMode.tsx
toddmedema: Done

app/components/base/QDLMode.tsx
toddmedema: updated

app/components/base/QDLMode.tsx
toddmedema: tested it, doesn't seem to do anything (the functions were copied and pasted from the markdown language definition). Removed and cleaned up this function

smartin015:
LG

PR #172, opened 2017-02-05
Cleanup SCSS

### LINKED ISSUE #160: Clean up SCSS
We should really eliminate all the magic numbers (by adding constants) and clean up the whole file.

smartin015:
I'd encourage a pull and exploratory test on this one... but our style code is now 50+ lines shorter even after moving the inline styles in QuestAppBar into styles.scss and is a lot cleaner.

Closes #160

### DIFF
--- a/app/components/ContextEditor.tsx
+++ b/app/components/ContextEditor.tsx
@@ -74,7 +74,10 @@ const ContextEditor = (props: ContextEditorProps): JSX.Element => {
   return (
     <div className="console">
       <div className="interactive">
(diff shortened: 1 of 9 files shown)


toddmedema:
@smartin015 looks great! Except that the splash page seems to be broken....
Specifically:
- on shorter screens, it doesn't scroll down, so the buttons are lost. Something about scrolling is broken?
- I'd put the h1 text and buttons above the video so that they don't get lost (since the video is so tall)
- try it in the mobile emulator - looks like the video isn't shrinking
- might be good to make the home page just a bit more inline with the website style, perhaps by making the new h1 text Cinzel font and a touch smaller

PR #179, opened 2017-02-11
add quest title to window title

toddmedema:


app/reducers/editor.tsx
```diff
@@ -23,6 +23,7 @@ export function editor(state: EditorState = defaultState, action: Redux.Action):
     case 'SET_LINE':
       return Object.assign({}, state, {line: (action as SetLineAction).line});
     case 'QUEST_RENDER':
+      window.document.title = 'Expedition Quest Creator - ' + (action as QuestRenderAction).qdl.getMeta()['title'];
```
smartin015: What does this do if the user doesn't set a title?

toddmedema: Just tried, doesn't break :) Title section just becomes an empty string

PR #182, opened 2017-02-13
Spellcheck!

toddmedema:
Looking for feedback on both code and UX

Closes https://github.com/Fabricate-IO/expedition-quest-creator/issues/56

screen shot 2017-02-13 at 10 09 05 am

Future improvements:
- add tests to spellcheck system?
- expose correct spelling(s) to user by clicking the misspelled word
  - allow users to click to apply the suggested spelling (make sure this change gets captured + saved)
  - allow users to ignore spellcheck suggestion (store in their local storage???)
- separate but related: DFS for validating op and ID names


smartin015:
Other improvements: 

* disable spellcheck on author, email, url, etc. quest attributes
* disable spellchecking anything inside an op node (as vars are probably all shortened names anyways).
* Allow users to ignore spellcheck suggestions. Are suggestions shown on hover, or does the user only know "this word is wrong"?

Is typo.js something we could list as a node dependency instead of a checked-in file?

toddmedema:
Improvements list for this PR:

[x] Typo.js as npm dependency
[ ] disable on quest attributes
[ ] disable on op nodes
[ ] disable on ID references - **goto ID** and (#id)
[ ] import / allow enemy names
[ ] rm html tags before checking

Future improvements:
- expose correct spelling(s) to user, either when hovering over word or gutter
  - allow users to click to apply the suggested spelling (make sure this change gets captured + saved)
- allow users to ignore spellcheck suggestion (store in their local storage???)

toddmedema:
@smartin015 updated with improvements listed above!

app/components/base/TextView.tsx
```diff
@@ -108,11 +127,11 @@ export default class TextView extends React.Component<TextViewProps, {}> {
       if (this.props.annotations) {
         session.setAnnotations(this.props.annotations);
 
-        for (var i = 0; i < this.props.annotations.length; i++) {
-          var line = this.props.annotations[i].row;
-          // TODO: Add and remove markers
-          //session.addMarker(new Range(line, 0, line, 1), 'ace_highlight-marker', 'fullLine');
-        }
+        // for (var i = 0; i < this.props.annotations.length; i++) {
```
smartin015: Yeah, this smells like junk code and we should remove it entirely. Pretty sure this can highlight whole lines, but the gutter annotations seem to work fine for now.

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
     );
   }
 }
+
+/* ===== Private helpers for spellchecking ===== */
```
smartin015: If these are private helpers, why aren't they defined as private methods within the TextView class?

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+
+function _loadDictionary(callback: any) {
+  const affPath = '/dictionaries/en_US_aff.txt';
+  const dicPath = '/dictionaries/en_US_dic.txt';
+  let affData: any = null;
+  let dicData: any = null;
+
+  $.get(dicPath, function(data) {
```
smartin015: Generally callouts to external data should be done within a redux action... but this is local and a hard requirement.

I feel like it'd be even simpler to just make the dictionaries into tsx files and require them. You could then even wrap them in the Typo() constructor call.

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+  const wordCount = words.length;
+  let pos = 0;
+  let bads: any[] = [];
+  for (let word = 0; word < wordCount; word++) {
+    const x = words[word] + '';
+    // skip ID references
+    if (x.indexOf('(#') === 0 && x.indexOf(')') > 0) {
+      continue;
+    }
+    const checkWord = x.replace(/[^a-zA-Z']/g, '');
```
smartin015: Add to REGEX constants?

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+    if (!dictionary.check(checkWord)) {
+      // skip valid encounter names
+      if (ENCOUNTERS.indexOf(checkWord.toLowerCase()) !== -1) {
+        continue;
+      }
+      bads[bads.length] = [pos, pos + words[word].length]; // equivalent to .push(), but (strangely) slightly faster
+    }
+    pos += words[word].length + 1;
```
smartin015: Isn't this incorrect positioning, since the line may have had several replacements of HTML tags and/or trigger(s)?

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+// Spell check the Ace editor contents.
+function _spellcheck(session: any, dictionary: any) {
+
+  if (!_contentsModified || _currentlySpellchecking) {
+    return;
+  }
+
+  _currentlySpellchecking = true;
+  for (let i in _markersPresent) {
+    session.removeMarker(_markersPresent[i]);
```
smartin015: Does removing markers before re-adding them cause flickering?

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+
+  try {
+    const lines = session.getDocument().getAllLines();
+    const lineCount = lines.length;
+    for (let i = 0; i < lineCount; i++) {
+      session.removeGutterDecoration(i, 'misspelled');
+      const line = lines[i];
+      const lineTrimmed = line.trim();
+      // skip metadata (only checks against first few lines for perf)
+      if (i === 0 || i < 20 && METADATA_FIELDS.indexOf(line.split(':')[0]) !== -1) {
```
smartin015: Does this not also skip the quest title?

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+        continue;
+      }
+      // skip op nodes
+      // TODO - check against ops that span multiple lines / skip all lines inside of op
```
smartin015: Thoughts on a new convention: something like `TODO(https://github.com/Fabricate-IO/expedition-quest-creator/issues/170)` that requires a bug to be assigned for every todo. This includes the full URL in case the TODO actually relies on another repo.

My concern is we have [30 TODOs](https://github.com/Fabricate-IO/expedition-quest-creator/search?p=2&q=TODO&utf8=%E2%9C%93) in our code, which is half again the number of open issues, and I feel like they aren't reflected in our work planning.

app/components/base/TextView.tsx
```diff
@@ -184,3 +203,98 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
+      if (lineTrimmed.indexOf('{{') === 0 && lineTrimmed.indexOf('}}') === lineTrimmed.length - 2) {
+        continue;
+      }
+      const misspellings = _misspelled(line, dictionary);
+      if (misspellings.length > 0) {
+        session.addGutterDecoration(i, 'misspelled');
+        for (let j in misspellings) {
```
smartin015: You could unnest this

app/constants.tsx
```diff
@@ -4,6 +4,16 @@ export const VERSION = packageJson.version;
 export const GITHUB_DOCS = 'https://github.com/Fabricate-IO/expedition-quest-ide/blob/master/docs/';
 export const DOCS_INDEX_URL = GITHUB_DOCS + 'index.md';
 export const DEV_CONTACT_URL = 'http://expeditiongame.com/contact';
+export const METADATA_FIELDS = [
```
smartin015: We should dedupe this against the [Normalize](https://github.com/Fabricate-IO/expedition-quest-creator/blob/master/app/parsing/validation/Normalize.tsx#L74) code. You could probably get this through `Object.keys(Normalize::questAttrs({}, null))`

webpack.config.js
```diff
@@ -16,7 +16,7 @@ const options = {
     ],
   },
   resolve: {
-    extensions: ['', '.js', '.ts', '.tsx', '.json']
+    extensions: ['', '.js', '.ts', '.tsx', '.json', '.txt']
```
smartin015: Make sure to remove these edits if you go for the require() approach to dictionaries.

app/components/base/TextView.tsx
toddmedema: 👍

app/components/base/TextView.tsx
toddmedema: Maybe? They are 60,000 lines of raw text, so adding any sort of formatting to them would notably increase file sizes. It'd also hold up the UI initialization for something that (IMHO) doesn't need to hold up the UI

toddmedema: Another way we could handle this, would be to have these dictionaries & their loading defined once as part of the main React.tsx file and its global loading (since they're global across text areas)

app/components/base/TextView.tsx
toddmedema: 👍

app/components/base/TextView.tsx
toddmedema: 👍

app/components/base/TextView.tsx
toddmedema: 👍

app/components/base/TextView.tsx
toddmedema: Nope, haven't seen any flickering (if I had to guess, because it's all in the same function, it might all happen before the next render happens)

app/components/base/TextView.tsx
toddmedema: could, though I thought it might it a bit clearer and faster, since this'll only happen if there are > 0 misspellings anyway?

app/constants.tsx
toddmedema: Moved to separate issue: https://github.com/Fabricate-IO/expedition-quest-creator/issues/183

app/components/base/TextView.tsx
```diff
@@ -82,15 +94,22 @@ export default class TextView extends React.Component<TextViewProps, {}> {
       // Must manually resize on re-render to account for SplitPane
       // adjusting the vertical height of Ace.
       ref.editor.resize();
+      const session = ref.editor.getSession();
 
-      var session = ref.editor.getSession();
+      if (!this.loadingSpellcheck) {
+        this.loadingSpellcheck = true;
```
toddmedema: Changed to `loadSpellcheck`, starts as true, set to false as soon as spellcheck starts loading

app/components/base/TextView.tsx
toddmedema: ...yep. Just improved!
screen shot 2017-02-18 at 1 43 25 pm

app/components/base/TextView.tsx
toddmedema: That makes sense! 
Fixed this specific todo:
screen shot 2017-02-18 at 2 14 48 pm

app/components/base/TextView.tsx
```diff
@@ -25,15 +33,111 @@ interface TextViewProps extends React.Props<any> {
...
+    let bads: any[] = [];
+    for (let i = 0; i < wordCount; i++) {
+      // remove non-word characters
+      const word = (words[i] + '').replace(REGEX.NOT_WORD, '');
+      if (!dictionary.check(word)) {
+        // skip valid encounter names
+        if (ENCOUNTERS.indexOf(word.toLowerCase()) !== -1) {
+          continue;
+        }
+        const pos = line.indexOf(word);
```
smartin015: If my suspicions are correct... this'll cause us to multi-highlight the first error word in a line and ignore all subsequent occurrences of the same word.

app/components/base/TextView.tsx
```diff
@@ -25,15 +33,111 @@ interface TextViewProps extends React.Props<any> {
...
+    return bads;
+  }
+
+  // Spellcheck the Ace editor contents.
+  private spellcheck(session: any, dictionary: any) {
+    if (!this.contentsModified || this.currentlySpellchecking) {
+      return;
+    }
+
+    this.currentlySpellchecking = true;
```
smartin015: drop "currently"?

app/components/base/TextView.tsx
```diff
@@ -25,15 +33,111 @@ interface TextViewProps extends React.Props<any> {
...
+        const line = lines[i];
+        const lineTrimmed = line.trim();
+        // skip title (first line) && metadata (only checks against first few lines for perf)
+        if (i === 0 || i < 20 && METADATA_FIELDS.indexOf(line.split(':')[0]) !== -1) {
+          continue;
+        }
+        // skip the inside of ops
+        let startOp = lineTrimmed.indexOf('{{');
+        let endOp = lineTrimmed.indexOf('}}');
+        if (startOp === 0) { // if line starts with an op, we're in an op
```
smartin015: This could be made a lot neater by a regex splitter. Match all text in multiple segments, and case based on the opening characters.

app/components/base/TextView.tsx
```diff
@@ -48,22 +152,23 @@ export default class TextView extends React.Component<TextViewProps, {}> {
...
   onRef(ref: any) {
-    if (this.ace && this.onSelectionChange) {
-      this.ace.editor.off('changeSelection', this.onSelectionChange);
+    const self = this;
+    if (self.ace && self.onSelectionChange) {
+      self.ace.editor.off('changeSelection', self.onSelectionChange);
     }
 
-    this.ace = ref;
+    self.ace = ref;
```
smartin015: Why switch to self here? Could instead use arrow syntax to improve the binding on events.

app/components/base/TextView.tsx
smartin015: I'd be down for doing this loading in the main file... I'm a bit leery of having web requests hidden in components.

app/components/base/TextView.tsx
toddmedema: 👍

app/components/base/TextView.tsx
toddmedema: Tried that, but got awkward quickly with double-nested functions, ie `setInterval(() => { this.spellcheck(session, dictionary); }, 500);`, where you'd have to call `.bind` twice. Since this happens multiple times in onRef, seemed cleaner to use `self`

toddmedema:
Note: tried writing tests for it, but it looks like getting tests to work requires Karma, which, since the Quest Creator and App are currently on different major versions of typescript, was proving to be a headache to implement / out of the scope of this PR, so dropped a separate issue: https://github.com/Fabricate-IO/expedition-quest-creator/issues/191

toddmedema:
@smartin015 updated! Much cleaner with the new implementation :)

app/components/base/TextView.tsx
smartin015: Let me direct you to: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_binding_of_this

PR #210, opened 2017-03-14
Add familyfriendly, /quest/rate and /quest/report

toddmedema:
Addresses part 1 of rate quests issue: https://github.com/ExpeditionRPG/expedition-app/issues/248

Pairs with https://github.com/ExpeditionRPG/expedition-app/pull/258

Also resulted in improved schema system. Had to move schemas to their own file to prevent circular dependencies in model schema generation.


toddmedema:
@smartin015 for re-review

models/quests.js
```diff
@@ -155,42 +109,11 @@ exports.search = function(userId, params, callback) {
...
-}
-
-function convertQuestXMLToMetadata(text) {
-  var $ = Cheerio.load(text);
-  return formatQuest($("quest"), {depth: 0});
+function convertQuestXMLToMetadata(text, callback) {
+  const $ = Cheerio.load(text);
+  const attribs = $("quest")[0].attribs;
+  delete attribs['data-line'];
+  Joi.validate(attribs, Schemas.quests, callback);
```
smartin015: Yesssss. This is nice.

models/migrations/1.2.0 - ratings.js
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE ratings (
```
toddmedema: fixed

models/query.js
```diff
@@ -79,7 +79,7 @@ exports.getId = function (table, id, callback) {
 // deletes row with matching id from table. Returns the id deleted.
 exports.deleteId = function (table, id, callback) {
 
-  exports.upsert(table, {id: id, tombstone: Date.now()}, (err, result) => {
+  exports.upsert(table, {id: id, tombstone: Date.now()}, 'id', (err, result) => {
```
toddmedema: it's the name of the column(s) to upsert against. Some tables have compound keys. Open to other ways to represent this :)

models/query.js
```diff
@@ -98,15 +98,18 @@ exports.create = function (table, object, callback) {
...
 
-// inserts object into table, updating it if ID already exists
-exports.upsert = function (table, object, callback) {
+// inserts object into table, updating it if [id] field already exists
+// if checking against a combination of fields, pass string 'field1, field2'
+exports.upsert = function (table, object, id, callback) {
 
   object = internals.objectJsToPg(object);
 
+console.log(object);
```
toddmedema: rm

toddmedema:
@smartin015 updated based on latest conversations, ready for review (again again again lol)

mail.js
```diff
@@ -4,25 +4,28 @@ const Config = require('./config');
...
-    text: text, // plaintext body
-    html: html, // html body
+    text: textMessage, // plaintext body
+    html: htmlMessage, // html body
   };
   if (process.env.NODE_ENV === 'dev') {
     console.log('DEV: email not sent (mocked). Email:');
     console.log('Subject: ' + subject);
-    console.log('Text: ' + text);
+    console.log('Text: ' + textMessage);
```
smartin015: I'd recommend logging htmlMessage here if we only want to log one of them.

models/migrations/1.2.0 - ratings.sql
```diff
@@ -0,0 +1,19 @@
...
+  text VARCHAR(2048),
+  email VARCHAR(255),
+  name VARCHAR(255),
+  difficulty VARCHAR(32),
+  platform VARCHAR(32),
+  players INT,
+  version VARCHAR(32)
+);
+
+ALTER TABLE quests ADD COLUMN familyfriendly BOOL;
```
smartin015: These `ALTER TABLE` queries can probs go away... I think rather than a "migrations" folder, we should just have a "sql" folder that has table definitions for all the things (instead of in header comments).

models/query.js
smartin015: I think changing the `id` param in `exports.upsert` below to e.g. `primaryKeyCSV` or some such would make it different from the "id" column here. Was just overwhelmed by ids everywhere. 

Really this seems like something that should be an attribute of the table, not of the upsert function call. But I guess that'd require a big change in how we do squel queries (per-table querying)

models/quests.js
```diff
@@ -265,3 +193,44 @@ exports.unpublish = function(userId, docId, callback) {
     return callback(err, id);
   });
 };
+
+exports.updateRatings = function(id, callback) {
+  Joi.validate(id, Schemas.quests.id, (err, id) => {
```
smartin015: Side note: I feel like our async SQL code would be waay nicer with [JS promises](http://odetocode.com/blogs/scott/archive/2015/10/01/javascript-promises-and-error-handling.aspx)

models/feedback.js
```diff
@@ -0,0 +1,81 @@
...
+const Squel = require('squel');
+
+const Mail = require('../mail');
+const Schemas = require('./schemas');
+const Query = require('./query');
+const Quests = require('./quests');
+const table = 'feedback';
+
+
+exports.getRatingsByQuestId = function(questId, callback) {
```
smartin015: `getFeedbackByQuestId`

models/feedback.js
```diff
@@ -0,0 +1,81 @@
...
+*/
+
+const Joi = require('joi');
+const Squel = require('squel');
+
+const Mail = require('../mail');
+const Schemas = require('./schemas');
+const Query = require('./query');
+const Quests = require('./quests');
+const table = 'feedback';
```
smartin015: `const FeedbackTable`? Because `table` made me go looking for function params.

models/quests.js
```diff
@@ -265,3 +193,44 @@ exports.unpublish = function(userId, docId, callback) {
...
+    if (err) {
+      return callback(err);
+    }
+
+    Query.getId(table, id, (err, quest) => {
+      if (err) {
+        return callback(err);
+      }
+
+      Feedback.getRatingsByQuestId(id, (err, ratings) => {
```
smartin015: After changing the method name... `(err, feedbacks)` or some such.

models/quests.js
```diff
@@ -265,3 +193,44 @@ exports.unpublish = function(userId, docId, callback) {
...
+        if (err) {
+          return callback(err);
+        }
+
+        const rates = ratings.map((rating) => {
+          return rating.rating;
+        }).filter((rating) => {
```
smartin015: Couldn't this be done faster in SQL land with a WHERE clause?

models/quests.js
```diff
@@ -265,3 +193,44 @@ exports.unpublish = function(userId, docId, callback) {
...
+          return (rating && rating > 0);
+        });
+        quest.rating_count = rates.length;
+        quest.rating_avg = rates.reduce((a, b) => { return a + b; }) / rates.length;
+
+        const maturities = ratings.map((rating) => {
+          return rating.maturity;
```
smartin015: Do we still have maturity fields in user ratings? I thought we removed them.

models/schemas.js
```diff
@@ -0,0 +1,89 @@
...
+const schemas = {};
+
+schemas.quests = {
+  id: ID_STRING, // <userId>_<docId>
+  publishedurl: Joi.string().uri().max(2048),
+  userid: ID_STRING,
+  author: NAME,
+  email: EMAIL,
+  familyfriendly: Joi.boolean(),
+  maxplayers: Joi.number().min(1).max(20),
```
smartin015: Suggest `ABS_MAX_PLAYERS` and `ABS_MIN_PLAYERS constants or something, since they're used in a bunch of places.

models/users.js
```diff
@@ -0,0 +1,32 @@
...
+  tombstone TIMESTAMP NULL DEFAULT NULL
+);
+*/
+
+const Joi = require('joi');
+
+const Query = require('./query');
+const Schemas = require('./schemas');
+
+const table = 'users';
```
smartin015: Again, when I see `table` it looks like it's defined in the function scope...

routes.js
```diff
@@ -4,31 +4,32 @@ const passport = require('passport');
...
-
-const HTML_REGEX = /<(\w|(\/\w))(.|\n)*?>/igm;
+// Phasing out as of 3/21/17; delete any time after 4/14/17
 // Joi validation: require title, author, email (valid email), feedback, players, difficulty
 // userEmail (valid email), platform, shareUserEmail (default false), version (number)
 router.post('/feedback', (req, res) => {
   const params = JSON.parse(req.body);
   // strip all HTML tags for protection, then replace newlines with br's
   const title = params.title.replace(HTML_REGEX, '');
+  const HTML_REGEX = /<(\w|(\/\w))(.|\n)*?>/igm;
```
smartin015: This is also defined in mail.js. Any chance we could dedupe them?

routes.js
```diff
@@ -142,4 +141,22 @@ router.post('/unpublish/:quest', (req, res) => {
...
+      if (err) {
+        throw new Error(err);
+      }
+      res.header('Access-Control-Allow-Origin', req.get('origin'));
+      res.header('Access-Control-Allow-Credentials', 'true');
+      res.end(id);
+    });
+  } catch (e) {
+    console.log(e);
+    return res.status(500).end(e.toString());
```
smartin015: I'd prefer a generic "something went wrong" to be shown to users, rather than them getting arbitrary exception text that could give away server state.

mail.js
```diff
@@ -4,25 +4,28 @@ const Config = require('./config');
...
 
 // TODO add template around message, including opt-out link
 // Then, track opt-outs in DB and check against opt-out list before sending message
 // Will need an opt-out route
 
 // to: single email string, or array of emails
-exports.send = function (to, subject, text, html, callback) {
+exports.send = function (to, subject, htmlMessage, callback) {
+  // for plaintext version, turn end of paragraphs into double newlines
+  const textMessage = htmlMessage.replace(/<\/p>/g, '\r\n\r\n').replace(HTML_REGEX, '');
```
toddmedema: This is the same code we were using in the old `/feedback` route, just moved it into the mail lib to make it more generic. Seemed to work fine in `/feedback` :)

mail.js
toddmedema: Good point

models/migrations/1.2.0 - ratings.sql
toddmedema: I've been doing migrations (on top of the definitions in each model folder) so that we get predictable results between dev and prod - ie these are the SQL commands I ran on dev to get it to work, so just run them on prod when you deploy this code. They're admittedly single-use files that aren't especially relevant after the version is deployed, but seems more open source / transparent / robust than having this in a text file on my laptop. LMK if that makes sense

models/quests.js
toddmedema: Added an issue: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/219

models/query.js
toddmedema: Oooh. Actually, that would make a lot of sense. We could make the local `const table = {name: 'feedback', id: ['id']}`, which would also make it easy to expand to future meta attributes!

Added a separate issue to prevent this PR from getting any more complex: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/220

models/quests.js
toddmedema: done

models/quests.js
toddmedema: good point; this is actually already done in `Feedback.getRatingsByQuestId`

models/quests.js
toddmedema: nope, removed. Done

models/feedback.js
toddmedema: Good idea; added to the model metadata issue: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/220

models/feedback.js
toddmedema: This is specifically get ratings (aka feedback with non-null ratings)

models/schemas.js
toddmedema: good idea, done

models/users.js
toddmedema: yep; added issue to improve the model metadata system: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/220

routes.js
toddmedema: If we upgrade the server-side code to use Typescript, we can pull it from `./app/regex` - issue added: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/221

routes.js
toddmedema: Done

toddmedema:
@smartin015 all comments address; added issues for some, questions on some, fixed some (noted in each)

mail.js
smartin015: "it worked before" doesn't preclude it not working later. But it seems like we're not doing anything crazy, so probs fine for now.

models/migrations/1.2.0 - ratings.sql
smartin015: Having a history of the schema and changes makes sense, but I think an "intended state" sql file is more useful to people. I propose we use something like https://github.com/joncrlsn/pgdiff to auto-generate these migration commands so we can regenerate any migration step based on the revision history if necessary.

routes.js
```diff
@@ -9,6 +9,7 @@ const Mail = require('./mail');
 const oauth2 = require('./lib/oauth2');
 const Quests = require('./models/quests');
 
+const GENERIC_ERROR_MESSAGE = 'Something went wrong. Please contact support.';
```
smartin015: Prefer a link to "support"

PR #213, opened 2017-03-16
fixing tests

toddmedema:
Looks like tests broke with the big upgrade due to missing ts lint modules. Added the modules, and now seem to be getting a ton of tslint errors, such as:

```
PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  SyntaxError: Unexpected token 'const'
  at app/spellcheck.test.tsx:1095
```

Which is weird, seems like it's not digging es6 syntax? Tried adding 'use strict' to the top of the file, didn't help. Any idea what's going on?


toddmedema:
Soooo.....I'm going to close this for now. This particular branch seems haunted. Master and all new PR's all compiling fine

PR #217, opened 2017-03-20
add scope._.views.<id> to docs

toddmedema:
Pairs with https://github.com/ExpeditionRPG/expedition-app/pull/267


docs/qdl_guide.md
```diff
@@ -29,12 +29,61 @@ Here we'll dive into more specific details about what's possible in QDL.
 
 Attributes can be added to specific cards and elements to give you additional capabilities.
 
+### IDs
+
 For **cards**, you can add an ID that can be jumped to from anywhere else in the quest via **goto elements**. Note: **IDs must be unique across the entire quest**. ID's should be written in camelCase and only use alphanumeric characters. This looks like:
 
 ```
 _roleyplay title (#id)_
```
smartin015: roleyplay?

docs/qdl_guide.md
```diff
@@ -29,12 +29,61 @@ Here we'll dive into more specific details about what's possible in QDL.
...
+```
+
+This can be very powerful - for example, if you wanted to create a conversation where the user could only ask each question once, you could do something like:
+
+```
+_Conversation loop (#conversation)_
+
+You prepare to ask John a question...
+
+* {{_.viewCount("question1") === 0}} Question 1
```
smartin015: Or even shorter... `{{!_.viewCount("question1")}}`

toddmedema: Nope :( MathJS doesn't like that. If we wanted a shorter / boolean check, we'd have to add another function, something like _.notViewed("question1") ; then we should probably add its inverse, _.viewed("question1") too

Issue #216, opened 2017-03-20
Add created timestamp to quests


toddmedema:
- change published to updated
  - will need to update app + api; should change published to updated for backwards compatibility
- that way we aren't destroying important metadata on new publish
- migration: copy updated values to created

PR #222, opened 2017-03-24
Add created field to quests

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/216


models/migrations/1.2.1 - quest created.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE quests ADD COLUMN created TIMESTAMP NULL DEFAULT NOW();
+
+UPDATE quests SET created = published;
```
toddmedema: @smartin015 note: leaving this file here b/c it's not a plain and simple creating a new column, there's also some actual data migration going on

PR #223, opened 2017-03-27
update + migrate quest ID to not include user id (stored in separate column)

toddmedema:
Well, this has been long overdue. Fixes the whole issue whereby the same quest document published by multiple people counts as different quests. But, we still (and always have) saved the user id as a separate column, so we could still incorporate user access controls in the future (ie let a friend edit, but don't give them power to publish)

Also, once quest ID's are fixed, we could do a bunch of fun stuff I was thinking about in the shower, including tracking user plays (quest id and timestamp), and showing them in search results when a quest has been updated since they last played (regardless of search filters, order, etc)


toddmedema:
Followup from IRL:

- some concern that google doc IDs aren't 100% unique / there may be collisions. We couldn't find any evidence of this online, but we did see some similarity in ID's of docs created by the same user, so there is a chance they aren't totally random (or incorporate the user ID as part of the doc ID). Unsure of how we could test this further, though.
- sharing settings also came up in discussion: it would be nice to allow quest creators to share with limited permission (ie read / write / publish). Read v write is doable with Drive share API, but publish would require more architecting since it would likely require us to maintain some shadow metadata on top of Drive, which isn't great...

smartin015:
An option for that second bullet point would be to require the most recent publisher to unpublish before others can publish. We can check against the list of shared users to make sure the most recent publisher is in the list of people that still have access to the doc... but of course that checking is all done client-side.

toddmedema:
That could be an interesting approach that'd require minimal code and UI. Just check if the user id of the person publishing matches the user id in the stored quest (and that the tombstone is null). If not, give them an error / alert that the existing owner must un-publish so that they can publish and become the owner. I like it, though it could be a bit of a hassle / security risk, in that, if you let someone else publish, they now have control and can prevent you (the rightful owner) from publishing...

BUT, it would allow a transfer of control / admin, which is another feature that usually exists in sharing situations.

Another possible approach: if anyone besides the original creator (stored in `userid`) tries to publish, the original creator is emailed with an "approve / deny" link. Publish events don't happen all that often, and emailing for permission would be an almost entirely UI-less solution (we'd have to let the guest publisher know that the creator has been emailed / the publish is pending their approval). That could even have a third link "make them the quest owner", which'd solve admin transfer issues...

Of course, then we'd also have to store the temporary publish data somewhere. Would be somewhat messy on the back-end, though that's preferable to a messy UI, IMHO

smartin015:
Security issues could be solved by un-sharing the document with that person and recovering ownership :) 

And as far as emails go... maybe we could just link to an "unpublish" action and let the person follow up manually or some such.

toddmedema:
1) hmm. So, the check for "is user X allowed to publish" would become something like:
  - are they the current user id && still shared on the document? then publish
  - are they shared on the document && the current user id is not shared on the document? then publish
  - otherwise, email the current user id for permission
2) makes sense

models/schemas.js
```diff
@@ -12,7 +12,7 @@ const MAX_PLAYERS = 20;
 const schemas = {};
 
 schemas.quests = {
-  id: ID_STRING, // <userId>_<docId>
+  id: ID_STRING,
```
smartin015: I'd recommend at least a comment that these follow google docs ids currently.

toddmedema:
@smartin015 well, I poked the wrong thing while doing final testing (tried to fix an error in Travis).... and now tests are broken again.

https://travis-ci.org/ExpeditionRPG/expedition-quest-creator

What I think is going on, is that the test files aren't getting linted or webpacked properyl, it's complaining about things like `SyntaxError: Unexpected token ')'`, but only with test files.

Thoughts?

smartin015:
I suspect this is actually because Travis is using phantomjs as a test browser/renderer and we're using chrome (https://github.com/ExpeditionRPG/expedition-quest-creator/blob/master/karma.conf.js#L38). Maybe change this to phantomjs and debug from there?

toddmedema:
You would think, but I'm also getting PhantomJS errors when running the tests locally:
```
PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  SyntaxError: Unexpected token ','
  at app/actions/dialogs.test.tsx:264


PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  SyntaxError: Unexpected token ')'
  at app/actions/editor.test.tsx:261


PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  SyntaxError: Unexpected token ')'
  at app/actions/quest.test.tsx:261


PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  SyntaxError: Unexpected token ')'
  at app/actions/user.test.tsx:261
```

Could you try pulling this branch, `npm install`, and see if `npm run test` works for you?

smartin015:
A couple fixes:

- Switched `npm run watch:test` to use PhantomJS so we get consistent errors
- Explicitly set babel loader to output es2015 since phantom doesn't know about arrow functions
- Removed unused sourcemap loader and prettified test style a bit.

Issue #228, opened 2017-04-02
Multipart form upload for quests


smartin015:
Large quests currently break when an upload is attempted. We should switch to a multipart upload to allow for mega-quests to exist.

toddmedema:
Looks like it's an issue with our body-parser lib; we should be able to increase the limit, but there seems to be a bug. Added an issue: https://github.com/expressjs/body-parser/issues/235

dougwilson:
Since you have https://github.com/ExpeditionRPG/expedition-quest-creator/blob/23b72809f36171f279ef637be975aacacd1f047f/app.js#L22 which accepts _every single type_, that's the only body parser that will ever get invoked. You can simply delete the two below it and add your limit to that once, since it's the only one that will do anything.

toddmedema:
@dougwilson thanks for the prompt answer!

Can confirm, that fixes it.

Question: is there a way to do all types EXCEPT one? In this case, we'd like it to default to text parsing, but in the special case of JSON, use `app.use(bodyParser.json({ type:'json/*' }));` - perhaps via the ability to use regexes for the type field?

Not breaking for our app, since it's (apparently, hah) been using just text parsing this entire time, but it does seem like that might be a useful addition?

dougwilson:
Just reverse the order of the app.use, so the more specific type is first.

PR #230, opened 2017-04-03
increase maximum quest upload size from 100kb to 5mb

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/228

Since this is a live bug & doesn't change existing features, deploying now, but worth continuing the discussion around getting body parser working properly


smartin015:
LGTM

PR #233, opened 2017-04-05
update app to latest version

toddmedema:
Updating as the refactors happen.

However, currently running into a Typescript error:

`Supplied parameters do not match any signature of call target.`
at `./app/components/QuestAppBarContainer.tsx:85:7 `

Which is 
`loadNode({autoRoll: false, numPlayers: 1, difficulty: 'NORMAL', showHelp: false, multitouch: false, vibration: false}, dispatch, newNode, ctx);`

And seems like it matches with
`export function loadNode(settings: SettingsType, dispatch: Redux.Dispatch<any>, node: XMLElement, ctx: QuestContext) {`

So not sure why it's breaking... #UnhelpfulErrorExplanation


smartin015:
Might be worth just trying another merge... a lot has changed on the app side.

toddmedema:
@smartin015 tried updating.... still running into the same errors. Did some more digging, and found myself a bit confused. It looks like loadNode no longer takes in context? How should the quest creator now pass in context when you hit "play from cursor" and have some default context info populated?

`export function loadNode(settings: SettingsType, dispatch: Redux.Dispatch<any>, node: ParserNode) {`

app/components/QuestAppBarContainer.tsx
```diff
@@ -80,9 +80,9 @@ const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): Quest
       }
 
       dispatch({type: 'REBOOT_APP'});
-      dispatch(initQuest(renderResult.getResult().children().eq(0), defaultQuestContext()));
+      dispatch(initQuest('0', renderResult.getResult().children().eq(0), defaultQuestContext()));
```
smartin015: Use `newNode` instead of the beginning of the quest here, then dispatch `loadNode` using the resulting `ParserNode` returned from `initQuest`

toddmedema:
@smartin015 fixed and ready for review!

app/components/QuestAppBarContainer.tsx
```diff
@@ -80,9 +82,9 @@ const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): Quest
       }
 
       dispatch({type: 'REBOOT_APP'});
-      dispatch(initQuest(renderResult.getResult().children().eq(0), defaultQuestContext()));
+      const result = dispatch(initQuest('0', questNode, ctx));
       // TODO: Make these settings configurable
```
smartin015: Add a bug for this TODO

PR #236, opened 2017-04-06
fix sending feedback to quest creator email + calculating rating avg on new rating

toddmedema:


mail.js
```diff
@@ -14,16 +14,18 @@ const HTML_REGEX = /<(\w|(\/\w))(.|\n)*?>/igm;
...
-    to: [].concat(to).join(','),
+    to: to,
     bcc: 'todd@fabricate.io',
     subject: subject,
     text: textMessage, // plaintext body
     html: htmlMessage, // html body
   };
   if (process.env.NODE_ENV === 'dev') {
     console.log('DEV: email not sent (mocked). Email:');
+    console.log('TO: '+ to);
```
smartin015: Why not leave the `to` field, and just do `mailOptions.to` here instead?

models/feedback.js
```diff
@@ -46,9 +46,9 @@ exports.submit = function(type, feedback, callback) {
         `;
 
         if (type === 'rating' && feedback.text && feedback.text.length > 0) {
-          Mail.send('expedition+questfeedback@fabricate.io', 'Quest rated ' + feedback.rating + '/5: ' + quest.title, htmlMessage, () => {});
+          Mail.send([quest.email, 'expedition+questfeedback@fabricate.io'], 'Quest rated ' + feedback.rating + '/5: ' + quest.title, htmlMessage, () => {});
         } else if (type === 'report') {
-          Mail.send('expedition+questfeedback@fabricate.io', 'Quest reported: ' + quest.title, htmlMessage, () => {});
+          Mail.send([quest.email, 'expedition+questreportedk@fabricate.io'], 'Quest reported: ' + quest.title, htmlMessage, () => {});
```
smartin015: `questreported`, minus the `k` :)

smartin015:
Also, any thoughts on how we could unit test this?

toddmedema:
@smartin015 hmm. We could mock out the actual send email function? `return transporter.sendMail(mailOptions, callback);` and check that it's properly called all the way from the top, ie when you hit the /rate or /report routes. Though, doing it all the way from the top would require tests to have access to a test db...

PR #253, opened 2017-04-19
Homepage refactor, fix login/logout

smartin015:
Previously, the `Sign Out` button wouldn't actually do anything (Because it's pretty much impossible to sign a user out that's authorized with google auth, aside from either logging them out of google or breaking the auth and re-asking permissions on next login). This is now fixed so that any navigation to a page without a document tag counts as a homepage view.

Also in this PR: A massive refactor of the homepage to match [expeditiongame.com](expeditiongame.com) visual style.


smartin015:
homepage

Issue #255, opened 2017-04-22
Notes


smartin015:
To help people better keep track of their ideas, we should allow for notetaking in a separate panel than the app. Some additional parsing could be done on the note text to inline help tooltips, spellcheck, and/or text completion. E.g for something like this:

```
Baron Badass: The protagonist of the quest. Enjoys cheese and kicking butt.
```

We could text-complete and spellcheck occurrences of `Baron Badass` in the quest, and set hovertext to the first bit of text.

Clicking defined words in the IDE could open the notes panel, and clicking it in the notes panel could step through instances of the word's use.

toddmedema:
Also thought: if you mention an ID in the notes, would be cool to make it a link to scroll the editor to that ID, a la https://github.com/ExpeditionRPG/expedition-quest-creator/issues/72

toddmedema:
As discussed, trimming down the spec on this issue to just be notes / no fancy dictionary or ID linking. We can always roll those out later, but no point in holding up an already-useful feature :)

smartin015:
Closing for now since the main feature is complete.

PR #256, opened 2017-04-22
Add notes feature in bottom tray

smartin015:
Add realtime API controlled, saveable user notes for quests. Already-created quests will have this feature added in on next load.

Part of #255.


smartin015:
notes

toddmedema:
- `bundle.js:1667 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the Main component.`
- probably out of scope here, but as we add more interactive UI components, would be good to hide the UI until it's ready
- thoughts on making the notes font minion pro? Would hopefully still fit with app theme, help differentiate it from quest text, and (as suggested from a user a while back), using a non-monospace font when possible would help it feel less intimidating to non-coders

Once that's fixed, looks good!

smartin015:
- The setState error is actually not related to this PR and also happens on master (it's a use-after-unmount issue in Main.tsx of expedition-app repo).
- Yes and no... we should hide or disable, but maintain sizing so the display doesn't shift around a bunch at load.
- As discussed, let's keep this font the same for now.

Issue #252, opened 2017-04-19
Dev speedup


toddmedema:
"You can also speed up babel-loader by as much as 2x by using the cacheDirectory option. This will cache transformations to the filesystem." -https://github.com/babel/babel-loader

PR #260, opened 2017-04-25
speed up dev with caching, rm unused package

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/252

Re-compiling after changing a file now takes milliseconds instead of ~a second


smartin015:
Neat. LGTM

Issue #242, opened 2017-04-12
Beginner Friendly: Play from cursor should support having cursor at the start


toddmedema:
You should be able to start the quest from the beginning by having the cursor at the top, then hitting "play from cursor"

If your cursor is before the first card, it should keep checking forward until it finds the first card, then play from there

PR #263, opened 2017-05-04
support play from cursor at start of quest

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/242


toddmedema:
@smartin015 just added tests, but running into weird issues actually running them... breaks on auth.tsx b/c global `utils` is not defined (external dependency, loaded via a JS script file). I tried:
- having utils = utils || {}, but typescript throws an error saying that you can't redefine something that's declared.
- using a karma pre-processor to inject utils as a global and, in karma loading a globals.js file before everything else that defined var utils = {}.... but neither had an impact
- changing the realtime-client script that provides utils to export it, and importing it into typescript rather than via HTML / the browser... but that broke random script things in the browser

... so, kinda stumped. But, at least there're now test stubs, and a few other testing improvements worked their way in (such as running babel on the app imports so that it stopped breaking on "undefined 'const'")

PR #267, opened 2017-05-15
switch to NPM GA module, track logins

toddmedema:
For post-push review


app/react.tsx
```diff
@@ -33,17 +34,27 @@ import {loginUser, setProfileMeta} from './actions/user'
...
   } catch (e) {
     console.log('Failed to parse anticipated Drive open URI: ' + window.location.search);
   }
 }
 
+ReactGA.initialize('UA-47408800-7');
+if (isQuest) {
+  ReactGA.pageview('/quest');
+} else {
+  ReactGA.pageview('/');
```
smartin015: This might be better off just by looking at `window.location.href`, since it's purely about which page is being viewed. The case were we failed to load the doc should probably set the URL properly anyways.

toddmedema: Hmm. It would be good to make `.href` more useful... right now, depending on if you pay attention to the hash or not, it's either always `/` (not differentiating between splash and editor), or is the quest ID (too unique for useful analytics). At some point we could consider using React Router for this sort of url-based conditional routing

PR #268, opened 2017-05-15
Publish / metadata form

toddmedema:
Overview:
- Removed attributes from XML, except title
- On load, if not defined, XML attributes are imported to GoogleDoc.metadata
- From there, the code pulls all metadata info from GoogleDoc.metadata
- added a parser error instructing users to delete attribute lines (see screenshot)
- validation of attributes is now in the UI / form, instead of the XML parser
- as a temporary compatibility measure, on publish, the metadata is inserted back into the XML. Once we've coded a version of the app that no longer relies on XML metadata, and we're 100% sure all users are on the new app, we can remove this stopgap code

Additional features (especially selecting genre, mood and content rating) will be rolled out separately given the large scope of this change.

screen shot 2017-05-15 at 6 24 36 pm
screen shot 2017-05-15 at 6 30 48 pm


app/actions/ActionTypes.tsx
```diff
@@ -65,6 +65,12 @@ export interface QuestRenderAction extends Redux.Action {
   msgs: LogMessageMap;
 }
 
+export interface QuestMetadataChangeAction extends Redux.Action {
```
smartin015: This does limit us to single value changes. Thoughts on making this a `SetQuestMetadataAction`?

app/actions/quest.tsx
```diff
@@ -103,71 +104,105 @@ function createDocNotes(model: any) {
   return string;
 }
 
-export function loadQuest(userid: string, dispatch: any, docid?: string) {
+function createDocMetadata(model: any, oldMeta: any) {
+  const map = model.createMap();
```
smartin015: Neat, so this should be realtime as well? Have you tried this with two different clients? What happens when:

- Client 1 publishes, with some changed values
- Client 2 clicks the publish button

Is publish state shown with the new values?

app/actions/quest.tsx
```diff
@@ -103,71 +104,105 @@ function createDocNotes(model: any) {
...
-    getPublishedQuestMeta(userid + '_' + docid, function(quest: QuestType) {
-      var xmlResult = renderXML(text);
-      quest = Object.assign(quest || {}, xmlResult.getMeta());
-      quest.id = docid;
-      quest.mdRealtime = md;
-      quest.notesRealtime = notes;
+
+    if (!metadata) { // Create metadata if it's an old quest w/o metadata attribute
+      // Default to any metadata set in the markdown metadata (migrate from the old format)
+      const oldMeta = renderXML(md.toString()).getMeta();
```
smartin015: What if there's a render error?

app/actions/quest.tsx
```diff
@@ -103,71 +104,105 @@ function createDocNotes(model: any) {
...
-    }
+export function publishQuestSetup(): ((dispatch: Redux.Dispatch<any>)=>any) {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    dispatch({type: 'QUEST_PUBLISHING_SETUP'} as QuestPublishingSetupAction);
+  }
+}
+
+export function publishQuest(quest: QuestType): ((dispatch: Redux.Dispatch<any>)=>any) {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    const xmlQdl = renderXML(quest.mdRealtime.getText());
```
smartin015: In my mind, `QDL` is the markdown language and `XML` is what it renders to. Can we rename this to `xmlRender` or some such, here and elsewhere?

app/actions/quest.tsx
```diff
@@ -103,71 +104,105 @@ function createDocNotes(model: any) {
...
+
+    // Insert metadata back into the XML. Temporary patch until ALL client apps
+    // are running a version that supports DB metadata
```
smartin015: Add a bug reference and TODO plz.

app/actions/quest.tsx
```diff
@@ -103,71 +104,105 @@ function createDocNotes(model: any) {
...
+        .attr('mintimeminutes', quest.mintimeminutes)
+        .attr('maxtimeminutes', quest.maxtimeminutes)
+        .toString();
+
+    dispatch({type: 'QUEST_RENDER', qdl: xmlQdl, msgs: xmlQdl.getFinalizedLogs()});
+    dispatch({type: 'REQUEST_QUEST_PUBLISH', quest} as RequestQuestPublishAction);
+    return $.ajax({
+      type: 'POST',
+      url: '/publish/' + quest.id,
+      data: xmlString,
```
smartin015: Seems like we still need to adjust the server endpoint to accept non-XML params as well, right?

app/components/DialogsContainer.tsx
```diff
@@ -19,9 +20,30 @@ const mapStateToProps = (state: AppState, ownProps: any): DialogsStateProps => {
...
       dispatch(setDialog(dialog, false));
     },
+    onRequestPublish: (quest: QuestType): void => {
+      Joi.validate(quest, {
+        title: Joi.string().max(100),
+        summary: Joi.string().max(200),
+        email: Joi.string().email(),
+        author: Joi.string().max(100),
+        minplayers: Joi.number().min(1).max(Joi.ref('maxplayers')),
+        maxplayers: Joi.number().min(Joi.ref('minplayers')).max(6),
```
smartin015: Any way we could simply stop people from setting the values improperly in the first place? Maybe move this validation to `handleMetadataChange`?

app/constants.tsx
```diff
@@ -8,30 +8,20 @@ export const METADATA_FIELDS = [
   'summary',
   'author',
   'email',
-  'url',
```
smartin015: Where did URL go?

app/components/DialogsContainer.tsx
```diff
@@ -19,9 +20,30 @@ const mapStateToProps = (state: AppState, ownProps: any): DialogsStateProps => {
...
+    handleMetadataChange: (quest: QuestType, key: string, value: any): void => {
+      dispatch(questMetadataChange(quest, key, value));
+    },
     onRequestClose: (dialog: DialogIDType): void => {
       dispatch(setDialog(dialog, false));
     },
+    onRequestPublish: (quest: QuestType): void => {
+      Joi.validate(quest, {
+        title: Joi.string().max(100),
+        summary: Joi.string().max(200),
```
smartin015: What about minimums for title, summary, author? Do these allow for empty fields?

app/parsing/render/BlockRenderer.tsx
```diff
@@ -244,7 +244,14 @@ export class BlockRenderer {
       let k = kv[0].toLowerCase();
       let v = kv[1].trim();
       attrs[k] = v;
+
+      if (k !== 'title') {
+        if (log) {
+          log.err('Quest attributes have migrated to the "Publish" button - simply delete this line.', '420', block.startLine + i);
```
smartin015: Needs an associated markdown file. Also are we sure these values are propagated to the publish section even if they're deleted before the user hits the publish button for the first time?

app/components/DialogsContainer.tsx
```diff
@@ -19,9 +20,30 @@ const mapStateToProps = (state: AppState, ownProps: any): DialogsStateProps => {
 
 const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): DialogsDispatchProps => {
   return {
+    handleMetadataChange: (quest: QuestType, key: string, value: any): void => {
+      dispatch(questMetadataChange(quest, key, value));
+    },
     onRequestClose: (dialog: DialogIDType): void => {
       dispatch(setDialog(dialog, false));
     },
+    onRequestPublish: (quest: QuestType): void => {
```
smartin015: Would be nice to see a couple unit tests to ensure that metadata changes & publishing validation are observed.

app/actions/ActionTypes.tsx
toddmedema: Hmm. That'd add bloat elsewhere, as each individual form element would then need to keep track of and pass on the entire metadata state

app/components/DialogsContainer.tsx
toddmedema: Correct values are mostly enforced by the UI itself (ie you can only select the numbers 1-6 for min and max players)

app/constants.tsx
toddmedema: Good point, worth a discussion. In migrating the metadata, I thought it'd make sense to remove the URL for two reasons:

1. Listing a URL in the Expedition app poses security threats - ie if users are clicking on author URLs, they're being directed to third party resources outside of our control.
2. External websites, in hindsight, seem like not a great addition to the Expedition experience. If the purpose of the URL is to provide more information about an author, I think it'd make more sense to have an account-level author bio (perhaps also moving name and email to account level) that gives authors a chance to tell users about themselves inside of the Expedition app (both a) eliminating the awkward app -> website -> app transitions and b) not requiring quest creators to host a website if they want to tell users anything about themselves).

Let me know if that makes sense / thoughts :)

app/actions/quest.tsx
toddmedema: Fixed!

app/actions/quest.tsx
toddmedema: What sort of render error are you thinking? Tried a couple different things like invalid metadata titles and funky indentation errors and it still caught the old metadata fine

app/actions/quest.tsx
toddmedema: renderXML -> xmlRender? Not sure what the difference there is. Changed xmlQdl to questObject for a bit more clarity, since it's not actually xml being returned (you have to call `.getResult()` to actually get the XML)

app/actions/quest.tsx
toddmedema: Added

app/actions/quest.tsx
toddmedema: Yep, but that'll be a separate PR. Trying to keep 'em more digestible :)

app/components/DialogsContainer.tsx
toddmedema: Cannot be empty - also added (very loose) minimums. Don't think we need to over-validate inputs unless we start seeing abuses :)

app/parsing/render/BlockRenderer.tsx
toddmedema: Yep - migration happens on quest load :) Added an error definition

Discovered that the error documentation and testing seems to have gotten disconnected, so added an issue for that - https://github.com/ExpeditionRPG/expedition-quest-creator/issues/271

app/components/DialogsContainer.tsx
toddmedema: Added some additional tests + stubs

app/actions/quest.tsx
smartin015: I was thinking if something throws an exception here, the error likely wouldn't be visible to the user. And it's not unlikely that the render code would fail in a mysterious way.

app/actions/ActionTypes.tsx
smartin015: After looking at the rest of the PR, this makes more sense. Let's leave it as is for now.

app/constants.tsx
smartin015: Seems fair. We'll probably end up with users posting URLs in their author bio areas... but that's probably fine for now.

app/actions/quest.tsx
smartin015: My comment was on `xmlQDL` being confusing. `questObject` is less confusing, but `Object` here adds no new information. 

What about `renderResult`?

app/actions/quest.tsx
smartin015: I don't see any TODO here.

app/components/DialogsContainer.tsx
smartin015: We should really be testing the publish flow here, particularly validation against bad values. Can you at least add some stubs for that?

app/components/DialogsContainer.tsx
smartin015: Loose minimums LGTM.

app/components/DialogsContainer.tsx
smartin015: Gotcha.

app/components/Dialogs.tsx
```diff
@@ -49,6 +52,111 @@ export class ErrorDialog extends React.Component<ErrorDialogProps, {}> {
...
+          value={metadata.get('email')}
+          floatingLabelText="Contact Email (private)"
+          onChange={(e: any, val: string) => { this.props.handleMetadataChange(this.props.quest, 'email', val)}}
+        />
+        <SelectField
+          floatingLabelText="Minimum Players"
+          value={metadata.get('minplayers')}
+          onChange={(e: any, i: number, val: number) => { this.props.handleMetadataChange(this.props.quest, 'minplayers', val)}}
+        >
+          <MenuItem value={1} primaryText="1" />
```
smartin015: Can we construct these based on the value in `constants.tsx` to prevent drift?

app/parsing/render/BlockRenderer.tsx
smartin015: Thanks :)

app/components/Dialogs.tsx
```diff
@@ -49,6 +52,111 @@ export class ErrorDialog extends React.Component<ErrorDialogProps, {}> {
...
+          label="Publish"
+          secondary={true}
+          onTouchTap={() => this.props.onRequestPublish(this.props.quest)}
+        />]}
+        titleClassName={'dialogTitle dialogGood'}
+        modal={false}
+        open={Boolean(this.props.open)}
+      >
+        <TextField
+          value={metadata.get('summary')}
```
smartin015: Min and max values based on Joi validation fields here?

app/actions/quest.tsx
toddmedema: Added a try-catch

app/actions/quest.tsx
toddmedema: Makes sense!

app/actions/quest.tsx
toddmedema: Added a URL, but good call, added `TODO` in front

app/components/DialogsContainer.tsx
toddmedema: Fair, added stubs

app/components/Dialogs.tsx
toddmedema: Good call, added

app/components/Dialogs.tsx
toddmedema: Material UI doesn't let you set text field max lengths in text fields, alas

smartin015: But it does have an `errorText` field: http://www.material-ui.com/#/components/text-field

We should be setting this (onBlur?) as the user fills out the form, instead of alerting them via popup after they try to submit. This can be a TODO.

app/components/Dialogs.tsx
```diff
@@ -66,6 +67,14 @@ export class PublishingDialog extends React.Component<PublishingDialogProps, {}>
     if (metadata === undefined) {
       return null;
     }
+    const playerItems = [];
+    for (let i = 1; i <= MAX_PLAYERS; i++) {
+      playerItems.push(<MenuItem value={i} primaryText={i} key={i} />);
+    }
+//     (new Array(MAX_PLAYERS)).map((val: number, index: number) => {
```
smartin015: Remove commented code

Issue #163, opened 2017-01-28
Preventing out-of-date editors


toddmedema:
Both for new features and important bugfixes, it'd be nice to prevent users from getting out of date.

- check every 24 hours against the github raw package.json
- parse out the version number, and if it's different
- show a persistent toast to the user that a new version is available with "refresh now" as the primary button and an "X" to close (it would re-appear every check interval until they refresh)

PR #272, opened 2017-05-19
add snackbar and new version notification

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/163

screen shot 2017-05-18 at 10 48 12 pm


app/actions/ActionTypes.tsx
```diff
@@ -98,3 +98,15 @@ export interface PanelToggleAction extends Redux.Action {
   type: 'PANEL_TOGGLE';
   panel: PanelType;
 }
+
+export interface SnackbarOpenAction extends Redux.Action {
+  type: 'SNACKBAR_OPEN';
+  message?: string;
+  action?: any; // function
```
smartin015: Type `() => any` ?

app/actions/ActionTypes.tsx
```diff
@@ -98,3 +98,15 @@ export interface PanelToggleAction extends Redux.Action {
...
+  actionLabel?: string;
+  timeout?: number;
```
smartin015: Let's not expose timeout until it's actually needed for some reason.

app/react.tsx
```diff
@@ -97,17 +99,34 @@ window.gapi.load('client,drive-realtime,drive-share', function() {
...
   }, 1000);
+
+  // every 12 hours, check for the latest version
+  setInterval(() => {
+    $.getJSON('https://raw.githubusercontent.com/ExpeditionRPG/expedition-quest-creator/master/package.json', (data) => {
+      if (data && data.version) {
+        const newVersion = data.version.split('.').map(Number);
+        const oldVersion = VERSION.split('.').map(Number);
+        if (newVersion[0] > oldVersion[0] || newVersion[1] > oldVersion[1] || newVersion[2] > oldVersion[2]) {
+          store.dispatch(openSnackbar('There\'s a new version of the Quest Creator available!',
```
smartin015: Do we really want this to have a timeout? The Inbox approach here is to show a notification with an `X` button.

Otherwise, this'll randomly show every 12 hours for a couple seconds. The user will probably miss it.

app/reducers/StateTypes.tsx
```diff
@@ -79,6 +79,14 @@ export interface DialogsState {
 
 export type ErrorsState = ErrorType[];
 
+export interface SnackbarState {
+  open: boolean;
+  message?: string;
+  action?: any; // function
```
smartin015: I'd prefer this being a part of `SetSnackbarAction` so we're not duplicating the interface in state and action. Also means we can get rid of `SnackbarCloseAction`.

app/reducers/snackbar.tsx
```diff
@@ -0,0 +1,27 @@
+import Redux from 'redux'
+import {SnackbarCloseAction, SnackbarOpenAction} from '../actions/ActionTypes'
+import {SnackbarState} from './StateTypes'
+
+const default_state: SnackbarState = {
+  open: false,
+  message: '',
+  timeout: 4000,
```
smartin015: Constant?

app/style.scss
```diff
@@ -243,6 +243,15 @@ body > :first-child:not(#react-app) { display: none; }
   }
 }
 
+.snackbar {
```
smartin015: Not sure the white background currently matches the theme of the IDE...

app/actions/ActionTypes.tsx
toddmedema: Ah

app/actions/ActionTypes.tsx
toddmedema: Fair

app/reducers/snackbar.tsx
toddmedema: Isn't this a constant? (we could expose default state if we wanted)... especially if we're no longer exposing timeout option, seems unneeded to move a number that's used in one place to a global list?

app/style.scss
toddmedema: Did it to stand out (it defaulted to black, and you couldn't see it)... open to suggestions :) (bright red?)

app/reducers/StateTypes.tsx
toddmedema: Not sure I follow... you're proposing that the interface for a new `SetSnackbarAction` would be = `SnackbarState`? And then places that open and close the snackbar would instead pass `true` or `false` as the open state? That seems reasonable / more code concise

app/react.tsx
toddmedema: Added an issue for Material UI lib to add a close button https://github.com/callemall/material-ui/issues/6913

toddmedema:
updated, migrated "publish success" dialog to toast, added a question

app/style.scss
smartin015: Maybe the dull red that we use for errors? Or just a darker grey shade.

app/reducers/snackbar.tsx
smartin015: It was being used in multiple places before, which suggested it should live as a const value in the file... probably fine now.

app/reducers/StateTypes.tsx
smartin015: Pretty much, although you'd still have to add a `type` field to the action. Can be done either by:

```markdown
export interface SetSnackbarAction extends SnackbarState {type: 'SET_SNACKBAR'}
```

or:

```markdown 
export interface SetSnackbarAction {
  type: 'SET_SNACKBAR'
  state: SnackbarState
}
```

The former probably involves less code, but we've done the latter elsewhere.

app/style.scss
toddmedema: Made it a darker shade of grey so that it's agnostic between good and bad - we can always colorize red / green later

app/reducers/StateTypes.tsx
toddmedema: Nice!

PR #275, opened 2017-06-02
Fix JSON exception on bad parse of bulleted text.

smartin015:
Originally reported by Greg Miller. Thanks, Greg!


toddmedema:
LGTM!

PR #277, opened 2017-06-03
Add genre and content rating to publishing

toddmedema:
screen shot 2017-06-02 at 11 49 32 pm
screen shot 2017-06-03 at 12 13 01 am


app/components/Dialogs.tsx
```diff
@@ -148,12 +157,34 @@ export class PublishingDialog extends React.Component<PublishingDialogProps, {}>
...
+        <SelectField
+          floatingLabelText="Content Rating"
+          value={metadata.get('contentrating')}
+          onChange={(e: any, i: number, val: number) => { this.props.handleMetadataChange(this.props.quest, 'contentrating', val)}}
+        >
+          {ratings}
+        </SelectField>
+        {metadata.get('contentrating') !== null && <div>"{metadata.get('contentrating')}" is defined as:
+          <ul>
+            <li>{rating.violence}</li>
```
smartin015: If we're not using the keys of the CONTENT_RATINGS const attributes, why isn't each rating just an array of strings? That way, we don't have to explicitly list them here and can just "for loop" 'em.

toddmedema: Was thinking that key'ing it provides a bit more structure to the data. Not much harder to loop over object attributes - added a loop to generate this :)

toddmedema:
Updated styling, generate rating description from loop.

As far as tests go, no tests are affected by this change - we have some stubs for quest publishing with valid and invalid metadata, but no tests to update just yet

screen shot 2017-06-03 at 8 45 53 am

PR #280, opened 2017-06-04
turn unpublish dialog into snackbar toast to match publish notification

toddmedema:
Small change, merging now for post review


smartin015:
LG

PR #282, opened 2017-06-06
Automatically share newly created quests with FabricateIO

toddmedema:
In case we need to debug or make changes

Note that the quests are discoverable in drive search, but don't automatically appear in the Fabricate drive - instead, you have to search them, add them to your drive, and THEN you'll be able to open it. That was the cause of the "Permission denied" bug I saw earlier tonight - it's just as true with quests you share with another person, until they add it to their drive, they'll get the same error.


app/actions/quest.tsx
```diff
@@ -77,19 +76,34 @@ export function newQuest(user: UserState, dispatch: any) {
...
+              method: 'POST',
+              params: {sendNotificationEmails: false},
+              body: {
+                role: 'writer',
+                type: 'domain',
+                domain: 'Fabricate.io',
+                allowFileDiscovery: true
+              }
+        });
+        request.then((json: any, raw: any) => {
```
smartin015: Couldn't you just do `window.gapi.client.request({...}).then(...)`?

app/actions/quest.tsx
```diff
@@ -77,19 +76,34 @@ export function newQuest(user: UserState, dispatch: any) {
...
+        }, (json: any) => {
+          console.log('Error connecting quest file to Fabricate.IO', json);
```
smartin015: Any chance we could log this to firebase/similar?

app/actions/quest.tsx
toddmedema: True!

app/actions/quest.tsx
toddmedema: Good idea - added some more logging of errors + user things to Google Analytics

PR #294, opened 2017-06-12
add engine and quest version metadata to quests

### LINKED ISSUE #289: Include qc version in quest publish metadata
While we're at it, would also be pretty easy to add an auto-incrementing version number for the quest itself

toddmedema:
Closes #289


toddmedema:
Updated to add second part, ability to set as a major release:
screen shot 2017-06-12 at 10 40 43 pm
Closs #289

app/components/base/Checkbox.test.tsx
```diff
@@ -0,0 +1,2 @@
+describe('Checkbox', () => {
```
smartin015: `it('blanks checkbox when value=false');`

`it('checks box when value=true');`

`it('triggers onChange when tapped');`

app/actions/Quest.tsx
```diff
@@ -220,7 +223,7 @@ export function publishQuestSetup(): ((dispatch: Redux.Dispatch<any>)=>any) {
 export function publishQuest(quest: QuestType): ((dispatch: Redux.Dispatch<any>)=>any) {
   return (dispatch: Redux.Dispatch<any>): any => {
     const renderResult = renderXML(quest.mdRealtime.getText());
-
+    const questVersion = (quest.questversion || 0) + 1;
```
smartin015: IMO, version is one of those things that should be handled and incremented server-side. Can we do this on the server instead?

toddmedema:
@smartin015 both comments addresses

app/actions/Quest.tsx
```diff
@@ -181,6 +182,7 @@ export function loadQuest(user: UserState, dispatch: any, docid?: string) {
       const xmlResult = renderXML(text);
       quest = Object.assign(quest || {}, xmlResult.getMeta(), {
         id: docid,
+        majorrelease: false,
```
smartin015: `majorRelease` to match style. Also, is this ever used apart from on POST to the server? Seems like it kind of clutters the attribute space.

toddmedema: @smartin015 updated; also, good point, switched it to be state-based + improved it to pass it via the request's querystring instead of embedding it in the XML (which is where we'll want to move the rest of the attributes that're currently ducttaped to the XML, once everyone's on the new version of the app)

PR #296, opened 2017-06-15
Actually useful quest crawler implementation

smartin015:
Currently detects invalid end states (no `**end**` trigger). More to follow.

- Moved stats generation from the original app PR
- Added skeleton `PlaytestPanel` which will show other helpful stats to the author
- Moved render/play logic out of the `QuestAppBarContainer` and into an action function
- Other small type fixes, updated app version


app/components/PlaytestPanel.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+export interface PlaytestPanelDispatchProps {}
+
+interface PlaytestPanelProps extends PlaytestPanelStateProps, PlaytestPanelDispatchProps {}
+
+const PlaytestPanel = (props: PlaytestPanelProps): JSX.Element => {
+  return (
+    <div className="console">
+      <div className="interactive">
+        TODO
```
toddmedema: Looks like results aren't displayed to the user?

smartin015: The panel isn't yet implemented/shown, but error results for cards are added to the gutter.

toddmedema:
Gotcha. LGTM as MVP

PR #297, opened 2017-06-15
Braintree API

toddmedema:
Merging now so that I can betatest iOS and Android apps, for post-review


routes.js
```diff
@@ -12,6 +13,13 @@ const Mail = require('./mail');
 const oauth2 = require('./lib/oauth2');
 const Quests = require('./models/quests');
 
+const gateway = Braintree.connect({
```
smartin015: Prefer `braintree` or similar, since `gateway` is a tad generic.

routes.js
```diff
@@ -195,4 +203,37 @@ router.post('/user/subscribe', (req, res) => {
...
+    res.send(response.clientToken);
+  });
+});
+
+
+router.post('/braintree/checkout', (req, res) => {
+  req.body = JSON.parse(req.body);
+  res.header('Access-Control-Allow-Origin', req.get('origin'));
+  res.header('Access-Control-Allow-Credentials', 'true');
+  gateway.transaction.sale({
```
smartin015: How will this record the user requesting the payment?

routes.js
toddmedema: Good point

routes.js
toddmedema: Ah, yep, just updated to capture customer and product info :)

Issue #185, opened 2017-02-18
Routes.js: Lock down CORS


toddmedema:
Important upgrade before payment system implemented

Issue #258, opened 2017-04-23
Use an ORM lib for data storage + separate dev from prod


smartin015:
*Migrated to expeditionrpg/expedition#296 by [spacetelescope/github-issues-import](https://github.com/spacetelescope/github-issues-import)*

If we use [sequelize](http://docs.sequelizejs.com/en/v3/) or something similar, we won't have to worry as hard about what data storage mechanism we use, and we'd actually be able to unit test things (since postgres doesn't allow ephemeral inmemory tables).

toddmedema:
Not pressing, but will make SQL testing easier - https://github.com/ExpeditionRPG/expedition-quest-creator/issues/184

toddmedema:
Once we do this, we can also block localhost from accessing prod

PR #301, opened 2017-06-19
limit CORS

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/185


routes.js
```diff
@@ -30,12 +31,19 @@ const GENERIC_ERROR_MESSAGE = 'Something went wrong. Please contact support by e
 const router = express.Router();
 router.use(oauth2.template);
 
+const blockCors = Cors({
+  credentials: true,
+  // allows expedition domains, localhost and file (for dev + mobile apps)
+  origin: /(expedition(game|rpg)\.com$)|(localhost(:[0-9]+)?$)|(^file:\/\/)/i,
```
smartin015: Can we go a bit further than this and conditionally exclude localhost if we're in a production environment?

routes.js
```diff
@@ -30,12 +31,19 @@ const GENERIC_ERROR_MESSAGE = 'Something went wrong. Please contact support by e
 const router = express.Router();
 router.use(oauth2.template);
 
+const blockCors = Cors({
```
smartin015: Prefer `limitCORS`, since we're not blocking them entirely.

routes.js
toddmedema: That means dev would never be able to test against the production server. Since we already have to allow file:// for mobile apps, I'm not sure that blocking localhost would add any additional protection

routes.js
toddmedema: Done

routes.js
smartin015: */me puts SRE hat on*

For a reliable service that significantly mutates DB state, dev should **never** touch prod or else there's a risk of data corruption/loss by rogue code (we are our own worst enemies). This is one reason why we should really move to an ORM so we can stop relying on dev'ing with prod data.

toddmedema: Gotcha, added this as a note to the ORM issue: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/258

PR #309, opened 2017-06-22
Prod fixes

### LINKED ISSUE #311: Safeguard production (deploy & rollback)
We had an almost 12 hour quest creator outage caused by a simple code bug that spiraled out of control due to deployment automation misconfiguration and failure in procedure.

See the [postmortem doc](https://docs.google.com/document/d/1SUpPxQua-tMB7TwMBcK0tXTlgjBQxg0gUzEMOLZnuRg/edit?usp=sharing) for followup action items.

smartin015:
Fixes #311


package.json
```diff
@@ -143,6 +142,7 @@
     "webpack-dev-server": "^2.4.2",
     "yargs": "^7.0.2"
   },
+  "devDependencies": {},
   "engines": {
     "node": "^6.9.0"
```
toddmedema: this should be node 8 now, correct?

package.json
```diff
@@ -82,9 +83,7 @@
     "typo-js": "^1.0.3",
     "warning": "^3.0.0",
     "webpack": "^2.2.1",
-    "winston": "^2.3.1"
-  },
-  "devDependencies": {
```
toddmedema: We can switch these back to being dev dependencies and not use heroku postbuild / resume our normal deploy process, correct? (unless we really don't care about the distinction between dev and prod dependencies)

smartin015: Done

package.json
smartin015: Done

smartin015:
Made a bunch of fixes (including docs) and set up master so that it only pushes to dev.  PTAL

webpack.static.config.js
```diff
@@ -0,0 +1,41 @@
+// Use this for building static error/maintenance pages.
```
toddmedema: Why is this separate from prod deployment config? Don't we want to always deploy these and decide whether to show them in the code?

smartin015: Fix'd. Note that we should eventually setup travis to autopush these pages to S3.

PR #315, opened 2017-06-26
Add explicit encounter tier JSON

### LINKED ISSUE #293: Catch misspelling of encounters that don't have explicit tier
User feedback has frequently mentioned this as a mistake. If the user doesn't explicitly specify their own tier, we should treat non-matching encounters as spelling errors and prevent publishing.

smartin015:
Includes compiler/playtest errors in case of improper JSON setting.

This is a sister PR to https://github.com/ExpeditionRPG/expedition-app/pull/371

Closes #293


errors/definitions/419.tsx
```diff
@@ -0,0 +1,18 @@
+export const NAME = `Detected a non-standard enemy without explicit tier JSON`;
+export const DESCRIPTION = `An enemy was listed that is not part of the standard deck of enemies, without a tier specified.`;
```
toddmedema: Might be helpful to suggest that users check spelling (does capitalization matter?) if they meant to include a standard enemy

app/parsing/crawler/PlaytestCrawler.test.tsx
```diff
@@ -30,6 +30,19 @@ describe('playtest', () => {
...
+      const msgs = playtestXMLResult(cheerio.load(`<quest>
+        <combat data-line="0">
+          <e>Custom Enemy</e>
+          <event on="win"><trigger>end</trigger></event>
+          <event on="lose"><trigger>end</trigger></event>
+        </combat>
+      </quest>`)('quest'));
+
+      expect(msgs.error.length).toEqual(1);
+      expect(msgs.error[0].text).toContain('without explicit tier');
```
toddmedema: Ah, error message testing with toContain make so much more sense / much more robust

errors/definitions/419.tsx
smartin015: Done.

smartin015:
Updated; PTAL

PR #316, opened 2017-06-26
Prod deploy fixes

toddmedema:
Fixes numerous small oddities in the current deploy / build scripts


toddmedema:
@smartin015 alright - prod's clean, the error and maintenance pages are all set up and deployed, and this PR's finally ready for review!

app/style.scss
```diff
@@ -23,8 +23,8 @@ $bgInteractive: rgb(20,20,20);
 $bgHighlight: #333;
 $textDisabled: #888;
 
-$bg_papyrus_img: url("/assets/img/papyrus-tiled.jpg");
-$bg_encounter_img: url("/assets/img/encounter-tiled.jpg");
+$bg_papyrus_img: url(../node_modules/expedition-app/app/images/papyrus-tiled.jpg);
+$bg_encounter_img: url(../node_modules/expedition-app/app/images/encounter-tiled.jpg);
```
smartin015: Aren't these still pulling from node_modules? I think this would be better done via files copied to dist.

webpack.dll.js
```diff
@@ -1,26 +1,27 @@
-var path = require("path");
-var webpack = require("webpack");
+// TODO update to work with latest node and webpack
```
smartin015: This'll probably change completely. IMO we're better off removing this file.

app/style.scss
toddmedema: Fixed

webpack.dll.js
toddmedema: RM'd

app/style.scss
```diff
@@ -307,7 +307,7 @@ label {
     }
 
     .app, button {
-      font-family: MinionPro, sans-serif !important;
+      font-family: Minion-Pro, sans-serif !important;
```
smartin015: Add quotes?

toddmedema: Yep, looks like consistent font name quoting is a good idea

Issue #298, opened 2017-06-15
Pay what you want: Stripe integration


toddmedema:
Stripe: https://stripe.com/docs/charges
- [x] don't need a token endpoint on our server, just a charge endpoint
- [x] /stripe/charge: create a charge and pass results back to app; add a metadata object to charge to include relevant info
- [ ] Set up Author authorization / account creation: https://stripe.com/docs/connect/express-accounts
  - [ ] `Authors` table. For now just google id, email and stripe id
    - [ ] Fetch + attach author info to quest.publisher (not quest.author, b/c already used...) on quest details fetch
      - [ ] Quest migration plan, ie what do we do with quests created pre-author table? Ideally switch to paying to Expedition directly / still allow PWYW. Also, after prod QC deploy, we can migrate the app to use quest.author as an object, but copy it from quest.publisher if that's being sent from the QC (then, in a month, have QC return as quest.author and remove that patch from app)
    - [ ] Using Express account type option, send destination account and amount: https://stripe.com/docs/connect/destination-charges
- [x] make sure API errors are getting handled properly, ie send a non-200 response back to app if something's wrong
- [ ] Final test w/ app + QC: https://github.com/ExpeditionRPG/expedition-app/issues/362
  - [ ] Can tip on a non-enabled quest; message changes to "Pay What You Want"
  - [ ] Can enable a quest to receive tips / register an author 
  - [ ] Can tip a quest, and that tip goes to the author w/ us taking a 10% cut

Future:
- [ ] Save payment: https://stripe.com/docs/sources/customers
  - [ ] First, figure out what the UI needs to let users select their payment source...
    - [ ] We'll probably need to send the UI the DB user object (incl stripeId) not just the Google Auth results
  - [ ] Create `user` db entry on login if it doesn't exist, or update lastLogin if it does
  - [ ] On charge, if !user.stripeId, call customers.create w/user.id and set it to the returned id if that succeeds, and THEN create the charge
  - [ ] Send user.stripeId with charges if it exists
  - [ ] Attach usage = reusable source to customers
- [ ] Switch from Express to Custom writer signup flow for more countries + more control (but more engineering effort)
- [ ] Recurring monthly user payments (Expedition Adventurer's Club)
- [ ] Support to split tips across multiple co-authors

Notes from previous integration attempt:

[...]

PR #318, opened 2017-06-29
Stripe integration

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/298


toddmedema:
Closing; now that we've split out API and QC, will have to rebuild this. Can still use this as a reference implementation

Issue #266, opened 2017-05-06
Refactor: Split out Quest Creator front-end from Expedition API, add serverside typescript + Hapi


toddmedema:
Allows QC to be hosted on CDN, will make it easier to convert the API part to typescript

Will have to host them in two separate places (api.expeditiongame.com?) - will need to leave quest creator running API as well for a while, to support older versions of the app / that don't point to the new URL

Big things to do post-refactor:
- Fix QC error definition generation + testing: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/271
- API-based announcement bar for app + QC: https://github.com/ExpeditionRPG/expedition-app/issues/370
- Make sure we're doing a good job of handling 500 errors and try/catching problems: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/186
- Make sure we're doing a good job of at least not breaking on non-latin1 characters (unit tests!!!) https://github.com/ExpeditionRPG/expedition-quest-creator/issues/157

smartin015:
I took a quick exploratory crack at getting the server running typescript. To use webpack to compile backend typescript, we can't use webpack-dev-server on the backend to host HMR files for development. The whole thing gets way simpler if the QC frontend is separate from the API server.

toddmedema:
Sounds about right. So, perhaps tomorrow morning we can set up that redirect so that we can do the separation in a month or so :-)

smartin015:
We're coming up against more and more things that are predicated on splitting this out.

smartin015:
We should use typescript server-side --^

PR #324, opened 2017-07-15
Split out commands for dev IDE and dev server

smartin015:
This requires both `npm run dev` and `npm run dev:ide` to be active when developing on the QC UI. In prod, `npm run start` still starts the app.js server and serves assets statically from `dist/`.

Also removed some unused GCP code.


smartin015:
Note: this is part of #266

PR #325, opened 2017-07-20
share doc on publish too

toddmedema:
Also fixes a bug where quest metadata wasn't being sent when publishing.

Tested locally, deployed to production and asked another writer to test it with one of their quests. Once that's confirmed, I'll send out asks to the remaining writers with older quests so that we can get them all shared :)


app/actions/Quest.tsx
```diff
@@ -238,14 +238,48 @@ export function publishQuest(quest: QuestType, majorRelease?: boolean): ((dispat
...
+      mintimeminutes: quest.mintimeminutes,
+      maxtimeminutes: quest.maxtimeminutes,
+      genre: quest.genre,
+      contentrating: quest.contentrating,
+      majorRelease,
+    });
     return $.ajax({
       type: 'POST',
-      url: '/publish/' + quest.id + '?majorRelease=' + majorRelease,
+      url: '/publish/' + quest.id + '?' + params,
```
smartin015: We'll need to be careful about this... max url length in chrome is ~2k characters. Probably better if we sent this metadata in the request body instead.

toddmedema: I think when I tried that, it led to massive bloat of the XML code itself because now it had to JSON encode the whole thing. Any ideas on how to get around that?

smartin015: Well, we could always insert the data as attributes in the `<quest>` element... \#dejavu

Or alternatively we could enforce a common format of post data, e.g. "everything up to the first newline is JSON, all following that is XML" or some such. This seems a tad gross though.

Third option would be to send a JSON blob, with a single attribute being the XML data as a string. Should only add a little extra overhead, and JSON parsing would likely not be too costly (for now).

toddmedema: For the sake of end users and since this is already on prod, merging.

But, thinking more about better future implementations...what if the Qc web app didn't upload the entire xml? What if it was rendered server side, and the Qc just transmitted the ID, then the server downloaded the Google Doc and parsed it using the same code? As a bonus, woukd prevent bad-xml injection attacks

smartin015: Seems legit, although there's the potential for version skew between the IDE and the server code. I guess that would just manifest as a "couldn't compile; please reload" error or some such.

Issue #157, opened 2017-01-26
Catch encoding errors, and support for non-latin1 characters


toddmedema:
Re-producible with smart quote characters: `’`

There are also many other useful characters outside of that character range (such as ✓)

Got
```
Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
    at i (http://quests.expeditiongame.com/bundle.js:17:25214)
```

Not only did it error, but that error didn't get passed to the saving indicator - just say there saying "saving..."

Possibility: encode when saving and loading: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa

Will need to make sure that the app is able to handle it properly (both in the quest editor, and after publishing)

smartin015:
&geq has worked fine in the past... there's also escaped code for other characters. Agreed that it would be nice though.

It sounds like this is actually two bugs - failure to catch the exception, and adding encoding support.

PR #328, opened 2017-07-28
fix latin character encoding

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/157

Confirmed on local and beta that it fixes the saving issue in the quest creator, and that it properly displays in the app for local dev + on publish (tested using a smart-quote character).

Currently blocking some polishes on old quests, and pretty sure at least one of our writers ran into this bug recently / got pretty frustrated, so will be good to have this fixed :)


toddmedema:
Hmm. Given that we can't provide useful instructions to the user on how to fix the issue, I'd prefer saving over mysterious failures. But I agree, mangling data also isn't fantastic. 

This (encoding issues) seems like a really common general issue. Are there other ways of solving this entirely? For example, replacing user input of all non-latin characters with that empty box character?

smartin015:
The real fix here is supporting unicode on upload... unsure how long this would take to implement though.

toddmedema:
It's not an upload issue, it's actually failing at the Google Drive API level. I did some looking through and could not uncover any way around this at the Drive API level, but didn't find anything. Another set of eyes would certainly help. We could also try switching to Drive API 3, though that could prove to break other things and won't necessarily fix this issue - "The differences between Drive API v3 and v2 are mainly changes to field names and removal of duplicate functionality." https://developers.google.com/drive/v3/web/migration

PR #329, opened 2017-07-28
skip flagging spelling 'errors' in elements we don't want to check

toddmedema:
For post-deploy review

Aka IDs, triggers, ops. Fewer false warnings = people paying more attention to real warnings :)

No measurable effect on performance (compared spellcheck execution times on local vs prod for same quest)


toddmedema:
Good call, added two stubs for it. Unsurprisingly, spellcheck testing isn't set up / doesn't work... hey Dave :D

PR #331, opened 2017-07-30
Separate backend server from quest creator code

### LINKED ISSUE #266: Refactor: Split out Quest Creator front-end from Expedition API, add serverside typescript + Hapi

smartin015:
A hermetic version of the api server is now available [HERE](https://github.com/ExpeditionRPG/expedition-api). This PR removes all of the duplicate code from the quest creator.

Figured I'd expose this to start conversation - should work just fine outside of docker, but I'm still working on a way to proxy requests appropriately within docker.

Fixes #266 - except for hapi integration which is tracked in https://github.com/ExpeditionRPG/expedition-api/issues/1


smartin015:
Side note: since this is a large (and probably breaking) change, we should roll this out to dev heroku + aws and test thoroughly before deploying to production.

toddmedema:
Agreed. We can deploy to betaapi.expeditiongame.com with a new test Heroku instance, and then point local copies of the app and quest creator UI against that to make sure everything's working as expected. We could then switch devquests to point to an S3 bucket hosting the static files that're rendered to look to betaapi.expedition, and see if that works as expected. For final prod deploy, we can update quests.expeditiongame.com to point to a cloudfront instance hosting static files that look to api.expedition (which is where the Heroku instance is) - we'll have to use Cloudfront so that we can have HTTPS (plus, files will serve faster :). That should all completely work fine - including the prod app. Then, we can update the prod heroku instance to build from the api repo, and if that works, we're in the clear!

toddmedema:
Note that ExpeditionGame.com is managed by my Hover account; I've added the Hover login to the shared company LastPass

smartin015:
Phase 1, complete! 

Hooked up dev to https://github.com/ExpeditionRPG/expedition-api and successfully tested against a local instance of the UI by running this PR with:

```
API_HOST=expedition-quest-creator-dev.herokuapp.com npm run dev
```

I also hopped into Heroku and set up a CNAME record for betaapi.*, but it didn't seem to do the right thing. Can you help troubleshoot this?

Once we've got that pointed to the dev instance, we can go ahead and host a static copy of this PR on AWS and point devquests.* at that to confirm everything works end-to-end.

toddmedema:
Make sure to push the update to HTTPS for prod webpack, then looks good to me! Nice work!

Issue #343, opened 2017-08-17
Error UX improvement: modal instead of separate link


smartin015:
Move error details from github markdown page to modal within the IDE, so that they don't have to leave the page to get more info (click the error icon in gutter to show the modal?)

Include a button for chat support on the error modal. (or at least mention chat button in bottom right)

PR #345, opened 2017-08-25
improved error button / jump to errors

toddmedema:
Addresses the first part of https://github.com/ExpeditionRPG/expedition-quest-creator/issues/343 by turning the old disabled "there are errors, good luck" button into an active button that you can click to automatically scroll to errors!

screen shot 2017-08-25 at 12 29 05 am


app/components/QuestAppBarContainer.tsx
```diff
@@ -60,6 +60,13 @@ const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): Quest
     onUserDialogRequest: (user: UserState) => {
       dispatch(logoutUser());
     },
+    onViewError: (annotations: AnnotationType[], editor: EditorState) => {
+      // Jump to the next error below the current line, looping back to the top error
+      const errors = annotations.filter((annotation) => { return annotation.type === 'error' });
+      const errorsAfterCursor = errors.filter((error) => { return error.row > editor.line }) || [];
+      const errorLine = (errorsAfterCursor.length > 0) ? errorsAfterCursor[0].row : errors[0].row;
```
smartin015: I don't think there's any requirements for the line annotations to be sorted... think we could add a quick call to sort() here so we for sure retrieve the next one down?

app/components/base/TextView.tsx
```diff
@@ -18,16 +18,19 @@ declare var gapi: any;
...
   realtimeModel: any;
-  annotations: AnnotationType[];
 
   // Use of SplitPane interferes with JS resize and rerendering.
   // When this value changes, the text view re-renders and the
   // correct vertical height is set.
   lastSizeChangeMillis: number;
+
+  // Hook for external control of scroll position
+  line?: number;
```
smartin015: Perhaps `lastTargetScrollLine` or similar? `line` indicates a current value, and there's a number of different things it could be at first glance (cursor line, top-of-window line, etc.)

PR #348, opened 2017-08-30
add encrypted s3 config to travis yml

toddmedema:


.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: node_js
+node_js:
```
smartin015: This shouldn't be necessary with the before_install command.

toddmedema: Seemed to be, when I looked at the Travis logs, it was installing Node v0.10. Might have installed another version of Node later on / after that, but seems inefficient to double-install node in every test instance

smartin015: That's what happens in `before_install` below - this is intentionally done to keep travis using the same node version as what's defined in package.json, since this has caused trouble for us in the past.

toddmedema: Gotcha

Issue #347, opened 2017-08-30
CI: Deploy to betaquests.expedition; clear prod CDN


toddmedema:
ON BETA: Hit beta API

ON PROD: Also fire CDN invalidation

Progress: https://github.com/ExpeditionRPG/expedition-quest-creator/pull/348

PR #350, opened 2017-08-30
Travis always deploy to beta; ./prod.sh deploy script

toddmedema:
closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/347


app/Constants.tsx
```diff
@@ -1,7 +1,7 @@
 const packageJson: any = require('../package.json');
 
 export const NODE_ENV = (process && process.env && process.env.NODE_ENV) || 'dev';
-export const API_HOST = (process && process.env && process.env.API_HOST) || '';
+export const API_HOST = (process && process.env && process.env.API_HOST) || 'http://betaquests.expeditiongame.com';
```
smartin015: Shouldn't this be `betaapi.*` now?

prod.sh
```diff
@@ -0,0 +1,24 @@
...
+
+read -p "Did you test on beta? (y/N) " -n 1
+echo
+if [[ $REPLY =~ ^[Yy]$ ]]; then
+  rm -rf dist
+
+  # Build the app - manually configure for production
+  export NODE_ENV='production'
+  export API_HOST='https://api.expeditiongame.com'
+  webpack --config ./webpack.dist.config.js
```
smartin015: No `-p` for optimization/production setting?

app/Constants.tsx
toddmedema: Ah, yep

prod.sh
toddmedema: Based on the webpack documentation, we're already doing both things that does (NODE ENV = production and trigger uglifyjs in the dist config); given that, I'd hesitate to add a new flag whose definition and effects may change in the future

smartin015: Fair 'nuff.

PR #365, opened 2017-09-10
Instruction style enforcement checks

### LINKED ISSUE #341: Instruction style enforcement (for later content detection)
Style enforcement of roleplay instructions (e.g. "draw x loot", "heal x health"). Dynamically detect when the user is likely to be doing these things and suggest a common style.

Besides being more consistent for players, also enables us in the future to detect / act on these instructions programmatically (both for crawler / balance detection, and for other quest/app implementations)

smartin015:
Also includes a minor cleanup of crawler validation code. This is done via the crawler so we can actually enforce instruction validity even if they're generated via op nodes!

Fixes #341

lootvalidation

### DIFF
--- a/app/playtest/PlaytestCrawler.tsx
+++ b/app/playtest/PlaytestCrawler.tsx
@@ -8,6 +8,17 @@ import {encounters} from 'expedition-app/app/Encounters'
 
 const cheerio: any = require('cheerio') as CheerioAPI;
 
+// Validators for instructions - these look at the preceeding 2 words
+// and expect a <verb> <count> <type> format, where <verb> is something like "gain" or "lose",
+// <count> is a positive integer, and <type> is any of ability/health/loot.
+// These matches are case insensitive and global, using the /gi suffix.
+const HEALTH_INSTRUCTION = /(\w+ \w+ (health|hp))/gi;
+const VALID_HEALTH_INSTRUCTION = /((gain|lose) \d+ health)/i;
+const ABILITY_INSTRUCTION = /(\w+ \w+ abili(ty|ties))/gi;
+const VALID_ABILITY_INSTRUCTION = /((learn|discard) \d+ abili(ty|ties))/i;
+const LOOT_INSTRUCTION = /(\w*\s*\w*\s*\w+ \w+ loot)/gi;
+const VALID_LOOT_INSTRUCTION = /((draw|discard) \d+ tier (I|II|III|IV|V) loot)|(discard \d+ loot)/i;
+
 // Surfaces errors, warnings, statistics, and other useful information
 // about particular states encountered during play through the quest, or
 // about the quest in aggregate.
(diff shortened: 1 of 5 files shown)


app/playtest/PlaytestCrawler.tsx
```diff
@@ -33,50 +44,85 @@ export class PlaytestCrawler extends StatsCrawler {
...
+  }
+
+  private verifyEnemyTiers(combatNode: Node<Context>, line: number) {
+    // Check that enemies are all valid or have tier overrides set.
+    combatNode.loopChildren((tag, child, orig) => {
+      if (tag !== 'e') {
+        return;
+      }
+      if (!encounters[child.text().toLowerCase()] && !child.attr('tier')) {
+        this.logger.err('Detected a non-standard enemy "' + child.text() + '" without explicit tier JSON', '419', line);
```
toddmedema: This might be worthy of a separate PR, but I think we should require non-standard enemies to also have health defined in their JSON

app/playtest/PlaytestCrawler.tsx
```diff
@@ -33,50 +44,85 @@ export class PlaytestCrawler extends StatsCrawler {
...
+        }
+      }
+      for (let m of (inst.match(ABILITY_INSTRUCTION) || [])) {
+        if (!m.match(VALID_ABILITY_INSTRUCTION)) {
+          this.logger.warn('Ability-affecting instructions should\nfollow the format "Learn/Discard <number> Abilit(y/ies)",\ninstead saw "' + m + '"', '434', line);
+        }
+      }
+      for (let m of (inst.match(LOOT_INSTRUCTION) || [])) {
+        if (!m.match(VALID_LOOT_INSTRUCTION)) {
+          this.logger.warn('Loot-affecting instructions should\nread as follows: "Draw 1 Tier IV Loot",\ninstead saw "' + m + '"', '434', line);
```
toddmedema: Draw (one/two/three/four/five/six)

app/playtest/PlaytestCrawler.tsx
```diff
@@ -33,50 +44,85 @@ export class PlaytestCrawler extends StatsCrawler {
...
+
+  private verifyInstructionFormat(roleplayNode: Node<Context>, line: number) {
+    roleplayNode.loopChildren((tag, child, orig) => {
+      if (tag !== 'instruction') {
+        return;
+      }
+      const inst = child.text();
+      for (let m of (inst.match(HEALTH_INSTRUCTION) || [])) {
+        if (!m.match(VALID_HEALTH_INSTRUCTION)) {
+          this.logger.warn('Health-affecting instructions should\nfollow the format "Gain/Lose <number> Health",\ninstead saw "' + m + '"', '434', line);
```
toddmedema: Should we also support "all" health? It's currently syntactic sugar to "Regain 12 health", but seems nicer to both quest writers and players

Issue #271, opened 2017-05-17
Fix errors/definitions generation/testing


toddmedema:
Not being tested against? Also, documentation not re-generating on test run, and there's no documentation on how to generate them

smartin015:
Karma is partially to blame here, since it runs from a browser and doesn't allow for FS operations. This'll be much more easily fixed once we've fully converted the server code to typescript and have testing for it.

PR #369, opened 2017-09-10
Annotation details modal

smartin015:
* Auto-loads from error definitions when gutter icon is clicked
* Includes suggestion to contact if blocked by errors
* Error loading was changed from using `fs` since this cannot be done in webpack/browser environment.
annotations

Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/271
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/343

### DIFF
--- a/app/components/Dialogs.tsx
+++ b/app/components/Dialogs.tsx
@@ -18,6 +18,7 @@ import {QuestType, ShareType, DialogsState, DialogIDType} from '../reducers/Stat
 import theme from '../Theme'
 import {MIN_PLAYERS, MAX_PLAYERS} from '../Constants'
 import {CONTENT_RATINGS, GENRES} from '../../node_modules/expedition-app/app/Constants'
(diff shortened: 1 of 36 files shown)


app/components/Dialogs.tsx
```diff
@@ -55,6 +56,72 @@ export class ErrorDialog extends React.Component<ErrorDialogProps, {}> {
...
+          <div>{goodExampleJSX}</div>
+          <strong>Invalid Examples</strong>
+          <div>{badExampleJSX}</div>
+        </div>
+       );
+    });
+
+    return (
+      <Dialog
+        title={'Annotation Details'}
```
toddmedema: This seems like a weird title, I doubt many users will think of these as annotations. Perhaps "Error details" or "Warning details" depending on if it's an error / warning?

PR #378, opened 2017-09-11
Playtest annotations... suggestions?

toddmedema:
Noticing an unfortunate bug, whereby typing in the editor resets the annotations (since that's how spellcheck currently does it). Which means as soon as you try to fix one playtest issue, all of them disappear.

This PR currently adds a "playtest on page load" which helps with the UX, but not much point in deploying this change till we figure out a better way to do the annotations in general.

Fun test quest (logged in as Fabricate) - 1,173 lines, takes 10 seconds to playtest:  http://localhost:8080/#0B64Jpu_POkg_NE5NNF9fSW1UYWM


app/actions/Quest.tsx
```diff
@@ -221,6 +221,8 @@ export function loadQuest(user: UserState, dispatch: any, docid?: string) {
       });
       dispatch(receiveQuestLoad(quest));
       dispatch({type: 'QUEST_RENDER', qdl: xmlResult, msgs: xmlResult.getFinalizedLogs()});
+      // Kick off a playtest after a moment so that slow computers can catch up if needed
+      setTimeout(() => dispatch(startPlaytestWorker(xmlResult.getResult())), 500);
```
smartin015: Perhaps we can try this even without the 500ms delay... my current theory is that calling the webworker in a reducer is somehow preventing repaint for some reason.

Barring that, we can do a setInterval() and make small advancements in the crawler code. Let's try the first option first though.

toddmedema: Fair; eliminated the additional delay.

I would be scared of doing any sort of set interval until we work out some of the performance bugs. We could also consider looping it into the same code that triggers the spellcheck (ie on every change to the text, if it's not already running, run it) but even that should wait until we figure out how to have the crawler stop locking up the page: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/379

PR #389, opened 2017-10-02
Remove line numbers from gutter

smartin015:
capture

The ace editor setting removes line numbers while still allowing error text. Intent here is to make this look less like a code editor and more like a word processor.


toddmedema:
UI updated to include line number and word count in bottom right: (had to move contact button up a bit, but seems fine):

code_example

Also updated homepage to have the new, more inviting screenshot

PR #395, opened 2017-10-05
show errors to writers

toddmedema:
Depends on https://github.com/ExpeditionRPG/expedition-qdl/pull/19 and https://github.com/ExpeditionRPG/expedition-app/pull/490

(so once that gets to deployed to master, we'll need to update package.json here)

Error displayed to user:

screen shot 2017-10-05 at 4 45 47 pm

I could also see showing this to the user as a dialog instead of a snackbar to give more space to the error message, include a "feel free to contact us if you're confused" and make it a bit more obvious that something is breaking. LMK what you think


app/React.tsx
```diff
@@ -25,6 +25,12 @@ declare var unescape: any;
 
 const Typo: any = require('typo-js');
 
+window.onerror = function(message: string, source: string, line: number) {
+  console.log('Error', message, source, line);
```
smartin015: Why not `console.error`?

app/React.tsx
```diff
@@ -25,6 +25,12 @@ declare var unescape: any;
 
 const Typo: any = require('typo-js');
 
+window.onerror = function(message: string, source: string, line: number) {
```
smartin015: Use `() => {}` style

app/React.tsx
```diff
@@ -25,6 +25,12 @@ declare var unescape: any;
...
+  console.log('Error', message, source, line);
+  store.dispatch(setSnackbar(true, 'Error! ' + message));
+  return true;
```
smartin015: What does `true` do here? Comment pls

app/Store.tsx
```diff
@@ -1,17 +1,21 @@
...
 declare var window:any;
 declare var require:any;
 declare var module:any;
 
-let devtools: any = window['devToolsExtension'] ? window['devToolsExtension']() : (f:any)=>f;
-let middleware = applyMiddleware(thunk);
+// from https://github.com/zalmoxisus/redux-devtools-extension#13-use-redux-devtools-extension-package-from-npm
+const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+const initialState = {preview: {}};
+const middleware = [thunk, getRemotePlayClient().createActionMiddleware()];
```
smartin015: `thunk` is no longer needed with the new remote play middleware.

app/React.tsx
toddmedema: Done

app/React.tsx
toddmedema: Good point, done

app/React.tsx
toddmedema: Added comment, prevents the firing of the default event handler

app/Store.tsx
toddmedema: Removed

PR #401, opened 2017-10-07
1.5.0 - fix scroll jump bug

toddmedema:


toddmedema:
LOL

To clarify / for documentation, it fixes scroll situations where the scroll position gets changed by something other than the user clicking on the text editor. For example, when CTRL + F searching, it would jump you to the next search result.... but the editor state doesn't receive the updated line # from this, and so the next time the state re-renders, it would push your cursor back to the last known line #. Made it much harder to search for things :)

Issue #386, opened 2017-09-28
Olark improvements


toddmedema:
Use Olark API to send quest title: https://www.olark.com/api#api.chat.onBeginConversation

Also, is it possible to include a link to a writing experience satisfaction survey after they message us? (And also let them know that we'll be in touch within 24 hours)

==== OLD

Specifically, for customer support chats, we're given the QC URL including the doc id hash, but we need the name to search it in Google Drive / add it to Drive so that we can open it in the QC

OR, we could support some sort of read-only mode since Fabricate has read-only access until it's added to Drive. Even if that mode just alerted with the name of the quest, would save a lot of time in asking "What's the name of your quest?" in every support request.

PR #402, opened 2017-10-08
olark: include quest name in message

toddmedema:
Waiting to hear back from Olark support - right now, the quest name is included if you're using their chat platform, but not in the offline email... would be awesome if we could include it in the offline email too

Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/386


toddmedema:
Heard back from support - unfortunately, you have to be logged in to receive this. So I might just leave my desktop logged in? Though that might give the false impression that someone is currently available to "chat" rather than that you're emailing support....

smartin015:
Is there an easy way to also include a link to the quest in the notification?

For email stuff, I wonder if we could hook into `mousedown` of the submit button and prepend it to the contents of `olark-form-message-input`?

toddmedema:
Tried a variety of methods, including hooking into the button press, `api.visitor.updateCustomFields` and `api.chat.updateVisitorNickname`, but none worked - so, sent another support request since sending a visitor's nickname with their email seems like a no-brainer...

PR #405, opened 2017-10-08
Re-enable quest crawler

### LINKED ISSUE #379: Crawler freezing page / poor UX
Try it on: https://quests.expeditiongame.com/#0B7K9abSH1xEOV3M2bTVMdWc4NVk

Any time you make a change and try to playtest it, the entire page freezes up for several seconds - which is annoying because you're actively trying to do something (playtest). We should either figure out a way to make it not freeze the UI (such as setTimeouts every ~100ms), or queue it to run the crawler on change, and then run it after there's been no interaction for a few seconds - and / or, display in the UI when the crawler is running, such as turning the "view errors" button into some sort of "checking for errors" spinner.

Also, it erases crawler errors as soon as you type something (spellcheck), so perhaps we should just disable it for now since that's a lot to fix?

### LINKED ISSUE #381: Buggy detection of syntax warnings
See https://quests.expeditiongame.com/#0B7K9abSH1xEOWVpEV1JGWDFtWmc

Primarily, had several warnings for "adventurer" over "player", but it just said `(in )`

smartin015:
This depends somewhat on fixes that snuck into https://github.com/ExpeditionRPG/expedition-app/pull/495. 

New worker code terminates existing workers if a new "play at cursor" button click fires, fixes infinite looping on ending work, work-shedding at max depth, and is somewhat more async to allow for possible yielding to other threads.

A dummy `mockContext` is currently needed due to some combat behavior; we may be able to remove this with better separation of business logic from the card templates.

Fixes #379 and fixes #381


app/React.tsx
```diff
@@ -20,6 +20,8 @@ declare var module: any;
 // For dev tools extension
 declare var window:any;
 
+window.onerror = (msg: string) => {console.error(msg);};
```
toddmedema: This is a duplicate of the window.onerror just a few lines below this

package.json
```diff
@@ -50,6 +50,7 @@
     "lodash": "^4.17.4",
     "markdown": "^0.5.0",
     "material-ui": "^0.17.4",
+    "mathjs": "^3.16.4",
```
toddmedema: Thought we didn't need this?

smartin015: Turns out we do, for ops work in the bottom panel.

PR #413, opened 2017-10-10
 fix bug where visually hiding line numbers broke error / annotation display

toddmedema:


toddmedema:
Closing; superior version here: https://github.com/ExpeditionRPG/expedition-quest-creator/pull/415

PR #414, opened 2017-10-11
Better auth: Pre-sign in

toddmedema:
Avoids popup blockers. Also allows for a bit of a better splash page UI, where we can indicate to the user that we do actually remember them (potentially enabling other useful buttons in the future like "Resume last quest")
screen shot 2017-10-10 at 8 51 31 pm


app/React.tsx
```diff
@@ -54,18 +53,24 @@ if (!window.location.hash && window.location.search.indexOf('ids') !== -1) {
...
 } else {
-  store.dispatch(setProfileMeta({loggedIn: false}));
   ReactGA.pageview('/');
 }
 
+// Try silently logging in
+// 10/10/2017: Also avoids popup blockers by making future login attempts
+// Trigger directly from the user action, rather than needing to load files
+window.gapi.load('client,drive-realtime,drive-share', () => {
+  store.dispatch(loginUser(false, questId));
```
smartin015: The `questId` param was removed from `loginUser`, right?

errors/errors.tsx
```diff
@@ -47,7 +48,7 @@ const errors: {[id: number]: ErrorType} = {
   431: e431,
   432: e432,
   433: e433,
-  434: e434,
```
smartin015: What about 434 though?

app/React.tsx
toddmedema: Nope, it's still there

PR #415, opened 2017-10-11
 fix annotation bug via css instead of forking

toddmedema:
screen shot 2017-10-10 at 9 54 14 pm


toddmedema:
Note that the tests actually pass, it's synk's security alert that's failing - most of which is from the QDL repo and is getting fixed here: https://github.com/ExpeditionRPG/expedition-qdl/pull/27

errors/errors.tsx
```diff
@@ -47,7 +48,7 @@ const errors: {[id: number]: ErrorType} = {
   431: e431,
   432: e432,
   433: e433,
-  434: e434,
```
smartin015: But... 434!

toddmedema: Wops - yeah, that'll get fixed when the other branch gets merged to master

Issue #173, opened 2017-02-05
Keyboard shortcut for play from cursor


toddmedema:

toddmedema:
Backlogging, not as convenient as it sounds since you still have to click in the app....

PR #416, opened 2017-10-11
ctrl + enter to play from cursor

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/173


app/React.tsx
```diff
@@ -77,14 +78,17 @@ window.onbeforeunload = function () {
 // Ctrl + <hotkey>
 window.addEventListener('keydown', function checkForCtrlS (event: any) {
```
smartin015: rename function

app/React.tsx
```diff
@@ -77,14 +78,17 @@ window.onbeforeunload = function () {
...
       case 's': // ctrl + s to save
         event.preventDefault();
-        const state = store.getState();
         if (state.editor.dirty) {
           store.dispatch(saveQuest(state.quest));
         }
         break;
+      case '\n':
+      case '\r':
+        store.dispatch(renderAndPlay(state.quest.mdRealtime.getText(), state.editor.line.number, state.editor.worker));
```
smartin015: Won't all this try to fire even if we don't have a loaded quest? Suggest gating it on presence of something loaded in `state.quest`.

Issue #419, opened 2017-10-11
Issue: Card Title icon "heart" not showing


ArcanaIreanol:
heart icon for theme (next to title card) is not working (same for roll, but d20 works)

not working: `_Test_ (#relations) {"icon":"heart"}`
not working: `_Test_ (#relations) {"icon":"hp"}`
not working: `_Test_ (#relations) {"icon":"roll"}`
working: `_Test_ (#relations) {"icon":"d20"}`

ArcanaIreanol:
same goes for:
* book (helper works)
* clock
* compass
* target

toddmedema:
Interesting! Since it's not breaking and would be a good first issue for someone looking to contribute, I've added a "help wanted" label - thanks for pointing this out :)

ArcanaIreanol:
I'd love to help, but I have to say I have no clue where to begin :p
could you point me where it is supposely set?

toddmedema:
I think you found the solution with https://github.com/ExpeditionRPG/expedition-app/pull/499 :)

Note that I added "closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/419" to your description for the PR - that'll make it automatically close this issue when the PR gets merged in!

PR #421, opened 2017-10-12
expedition-quest-creator #419 - add missing icons in .md

ArcanaIreanol:


toddmedema:
@ArcanaIreanol looks great! Thank you :)

PR #432, opened 2017-10-30
fix error definition testing; progress towards spellcheck testin

toddmedema:


app/dictionaries/en_US_combined.test.tsx
```diff
@@ -0,0 +1 @@
+// Filler
```
smartin015: eh?

toddmedema: It needed a file, but there's nothing to test

PR #442, opened 2017-11-11
allow toggling of line numbers by clicking the lines button in bottom right

toddmedema:
For power users only, but have heard that they liked it / miss it :) (still disabled by default)

In the future, could store and recover this setting from cookies


toddmedema:
Merging / for post review

PR #443, opened 2017-11-14
fix line skipping on delete

toddmedema:


toddmedema:
Merging / for post review

Issue #380, opened 2017-09-20
Error to user when using a Horror enemy & Horror not set to required


toddmedema:
Note: important race condition: If you show an error, users can't publish, which means they can't change this setting. So either the error needs to provide a way to change it (complicated code + mediocre UX) or it should be a warning instead of an error (which won't prevent people from publishing broken quests)

toddmedema:
Pre-requisite: Pass settings to the crawler (for now: `expansionhorror` flag; in the future, could pass a duration flag for longer validation / pre-publish runs)

UI thought: similar to the issue here: https://github.com/ExpeditionRPG/expedition-quest-creator/issues/444 - might warrant splitting out publish vs validate buttons? OR "Publish" and "Quest Settings", so that you can change the settings at any time, independent of publishing (which is also a bit clearer).

Issue #444, opened 2017-11-14
Publish button / dialog: Final validation before publish if changed since last


toddmedema:
In the same way we keep track of isDirty, we should also keep track of isValid - YES | ERRORS | UNKNOWN - and use that to inform the publish button state. Can show a green checkmark if valid (which normal white text). The real question, if the state is unknown, what should happen? Should clicking "publish" run the validator? (disconnects button label from function) or should the button label change? (where'd the "publish" button go?)

^^ thought: have an "errors / validate" button next to publish button to capture the now multitude of states? OR, have this final validate hook into the publish dialog, allowing you to make changes to the metadata but not actually click "publish" until it's finally confirmed good?

toddmedema:
Another approach: Keep validation the way it currently works, but move the settings adjustment stuff to a separate "Quest Settings" button / dialog - since, like an Android app, one of those buttons is really release management, and the other controls the store listing.

That being said, this could result in confusion around what it means to save settings - since they wouldn't update in the app. But, that confusion already exists to a degree since settings are auto-saved, but still only pushed to the app on a publish.

OR, we could also allow users to change the settings / listing without publishing new XML....

toddmedema:
Removing from the backlog since it appears possible to accidentally game the system and publish an invalid quest if you're able to hit "publish" before the validator re-runs

PR #447, opened 2017-11-26
add crawler settings; give warning on horror enemy w/o horror enabled

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/380 (note that it would be better for this to be an error, but there's a lot of other UI stuff required for that to not result in a race condition, see https://github.com/ExpeditionRPG/expedition-quest-creator/issues/444)


app/playtest/PlaytestCrawler.test.tsx
```diff
@@ -7,7 +8,7 @@ var expect: any = require('expect');
 const cheerio: any = require('cheerio') as CheerioAPI;
 
 function playtestXMLResult(elem: Cheerio): LogMessageMap {
-  const crawler = new PlaytestCrawler(null);
+  const crawler = new PlaytestCrawler(null, {} as PlaytestSettings);
```
smartin015: Why does this need a type assertion? Can't we just make the settings attributes optional?

app/playtest/PlaytestCrawler.tsx
```diff
@@ -80,15 +83,23 @@ export class PlaytestCrawler extends StatsCrawler {
...
-    // Check that enemies are all valid or have tier overrides set.
+  private verifyEnemies(combatNode: Node<Context>, line: number) {
+
     combatNode.loopChildren((tag, child, orig) => {
       if (tag !== 'e') {
         return;
       }
-      if (!encounters[child.text().toLowerCase()] && !child.attr('tier')) {
+      const encounter = encounters[child.text().toLowerCase()];
+      // Check that enemies are all valid or have tier overrides set.
```
smartin015: `valid` => `standard`?

app/playtest/PlaytestCrawler.tsx
```diff
@@ -80,15 +83,23 @@ export class PlaytestCrawler extends StatsCrawler {
...
+      if (!encounter && !child.attr('tier')) {
         this.logger.err('Detected a non-standard enemy "' + child.text() + '" without explicit tier JSON', '419', line);
       }
+      // Check that defined, non-base enemies listed do not belong to disabled content sets
+      if (encounter && encounter.set !== 'base') {
+        if (!this.settings['expansion' + encounter.set]) {
+          this.logger.warn('Detected a Horror enemy (' + child.text() + ') but this quest does not require The Horror.', '437', line);
```
smartin015: Looks like we should change `Horror` here in both sections to `encounter.set`

PR #469, opened 2018-02-09
display loot points in UI

toddmedema:
screen shot 2018-02-08 at 21 02 25

screen shot 2018-02-11 at 10 04 20


app/actions/User.tsx
```diff
@@ -34,12 +34,20 @@ export function loginUser(showPrompt: boolean, quest?: boolean | string): ((disp
...
+                      id: data,
+                    };
+                  }
                   const user = {
                     loggedIn: true,
-                    id: data,
                     displayName: googleUser.name,
                     image: googleUser.image,
                     email: googleUser.email,
+                    ...data,
```
smartin015: Rather than spreading a JSON object, I think it'd be better to explicitly assign id and lootPoints. That way we avoid initial assignment bugs (like the current one where data is only assigned the id attribute and not lootPoints when we catch an error in parsing JSON).

PR #471, opened 2018-02-18
fix bug with QDL and [art] detetion

toddmedema:


app/playtest/PlaytestCrawler.tsx
```diff
@@ -77,6 +77,9 @@ export class PlaytestCrawler extends StatsCrawler {
 
   private verifyRoleplayArt(roleplayNode: Node<Context>, line: number) {
     roleplayNode.loopChildren((tag, child, orig) => {
+      if (tag === 'choice') {
+        return; // bug with loop children returning all children text dumped into a single line
```
smartin015: If this is a bug, is there a bug reference in the QDL repo? if so, link it here as a TODO. If not... suggest clarifying this (I'm not sure what this is fixing).

toddmedema: Added link to the bug; let me know if you have any thoughts on how to address it in a more elegant manner.

https://github.com/ExpeditionRPG/expedition-qdl/issues/63

smartin015: I'm not convinced this is a bug... see comments on the issue.

app/playtest/PlaytestCrawler.tsx
```diff
@@ -78,7 +78,9 @@ export class PlaytestCrawler extends StatsCrawler {
   private verifyRoleplayArt(roleplayNode: Node<Context>, line: number) {
     roleplayNode.loopChildren((tag, child, orig) => {
       if (tag === 'choice') {
-        return; // bug with loop children returning all children text dumped into a single line
+        // bug with loop children returning all children text dumped into a single line
+        // https://github.com/ExpeditionRPG/expedition-qdl/issues/63
```
smartin015: See response to the issue... your PR is the correct thing to do, and it's not a bug that loopChildren does what it does. Let's remove these comments, then LGTM.

PR #473, opened 2018-03-09
Spell Error Fix

h-ckett:
Line 49, changed "jump to the nest result" to "jump to the next result".


toddmedema:
Thanks for the fix, and good catch!

Issue #477, opened 2018-04-14
Limit simulator functionality


toddmedema:
block simulator from doing anything except play from cursor, perhaps show a "looking for the web version of the app? Head over to app.expeditiongame.com. Trying to play your quest? Hit "Play from cursor" in the top right." when they try to do something else

Had to service a surprising number of support requests of people who publish their quest then play it in the simulator...

toddmedema:
This could also be achieved with an overlay on load that disappears once users click "Play from Cursor" for the first time

PR #478, opened 2018-04-15
add tutorial overlay to app simulator instructing the use of play from cursor

toddmedema:
screen shot 2018-04-15 at 09 24 35

closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/477


app/Style.scss
```diff
@@ -427,6 +428,24 @@ body .olark-launch-button { /*must be more specifc than olark styles */
     .app, button {
       font-family: $fontFamilyBody !important;
     }
+
+    &.play-from-cursor-tutorial {
+      &:after {
+        content: 'Click "Play from Cursor" above \a to test your quest';
```
smartin015: I don't usually expect content to exist in CSS configs. Would you mind making this explicit JSX in `QuestIDE.tsx`?

Also, it's missing punctuation at the end of the sentence.

toddmedema: Ah yep, that's a much more sensible way to do it

PR #479, opened 2018-04-15
use beta credentials by default

toddmedema:
To get rid of the security warning on our production integration, we need to have a separate credential for development / non-secure URLs. Confirmed working on my local instance, thinking we can then upgrade your dev setup when you get home, then beta, then prod.


app/Auth.tsx
```diff
@@ -1,8 +1,10 @@
 declare var utils: any;
 declare var window: any;
 
+const Config = require('../config.json');
+
 export const realtimeUtils = new utils.RealtimeUtils({
-  clientId: '545484140970-r95j0rmo8q1mefo0pko6l3v6p4s771ul.apps.googleusercontent.com',
+  clientId: process.env.OAUTH2_CLIENT_ID || Config.OAUTH2_CLIENT_ID,
```
smartin015: Thoughts on instead using nconf like in [expedition-api](https://github.com/ExpeditionRPG/expedition-api/blob/master/app/config.ts)? That way we could have a standard environment setting across repos.

toddmedema: Incidentally, not possible due to a bug in nconf that you reported a year ago... https://github.com/indexzero/nconf/pull/252

